PR #20, opened 2019-12-11
Update ARCH and BITS parser

sophia-guo:
Update ARCH and BITS parser to work with SPEC linux_aarch64. So
JVM_OPTION could be set properly. 

Signed-off-by: Sophia Guo <sophiag@ca.ibm.com>


sophia-guo:
related with https://github.com/AdoptOpenJDK/openjdk-tests/issues/1503

sophia-guo:
Tested locally as AdoptOpenJDK do not have machine enable for a jenkins job. 
Other platforms are tested with internal jenkins job.

PR #31, opened 2020-01-20
Correct dependency curl processing on zOS

andrew-m-leonard:
Signed-off-by: andrew-m-leonard <andrew_m_leonard@uk.ibm.com>

### DIFF
--- a/scripts/getDependencies.pl
+++ b/scripts/getDependencies.pl
@@ -232,7 +232,14 @@ sub downloadFile {
 	my $output;
 	# ToDo: should use curl only
 	if ($os eq 'os.zos') {
-		$output = qx{curl -k -o $filename $url 2>&1};	
+                # Delete existing file in case it is tagged incorrectly which curl would then honour..
+                qx(rm $filename);
+                # .txt SHA files are in ISO8859-1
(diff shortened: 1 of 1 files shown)


andrew-m-leonard:
On zOS curl processes existing files according to how they are already tagged.
The TKG get.testDependency pipeline stage copies the pre-built artifacts from job test.dependency using copyArtifacts and all files are tagged on zOS as IBM-1047.
This unfortunately means when the later getDependency.pl script runs, when it comes to asmtools.jar.sha.txt it can't read it because it thinks it's IBM1047 when in fact its ISO8859-1. It then tries to download the asmtools.jar again, but because that is also tagged IBM1047, curl thinks it should convert the downloaded jar into IBM1047, hence the corruption and wrong sha...

This fix removes the previous downloaded file before calling curl, and also encodes it correctly, ISO8859-1 for the sha.txt, and untagged for the .jar.

andrew-m-leonard:
sanity.functional Grinder_TKG running successfully with this PR: https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder_TKG/148/console

PR #37, opened 2020-02-06
Using TEST_JDK_HOME/java to launch ant for TKG compilation

renfeiw:
- move JAVA_HOME export to makefile

Signed-off-by: Renfei Wang <renfeiw@ca.ibm.com>


llxia:
To keep the test story simple, we would like to use the same java to launch ant for both TKG compilation and test material compilation.

PR #41, opened 2020-02-14
Count tests skipped at runtime as passing

ymanton:
TestNG returns a specific exit code if all tests are skipped at
runtime. For such cases we should count the test as passing
because (presumably) the test knows what it's doing.

Signed-off-by: Younes Manton <ymanton@ca.ibm.com>

### DIFF
--- a/settings.mk
+++ b/settings.mk
@@ -83,7 +83,7 @@ ifndef JVM_VERSION
 
 ifeq (hotspot, $(JDK_IMPL))
 	JVM_VERSION = $(OPENJDK_VERSION)
-else 
+else
 	JVM_VERSION = $(OPENJDK_VERSION)-$(JDK_IMPL)
 endif
 
@@ -94,7 +94,7 @@ ifneq (,$(findstring win,$(SPEC)))
 P=;
 D=\\
 EXECUTABLE_SUFFIX=.exe
-RUN_SCRIPT="cmd /c" 
+RUN_SCRIPT="cmd /c"
 RUN_SCRIPT_STRING=$(RUN_SCRIPT)
 SCRIPT_SUFFIX=.bat
 PROPS_DIR=props_win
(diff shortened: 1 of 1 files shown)


smlambert:
I believe test count is for counting test targets, NOT specific testcases within the target.  As such, I believe it is inappropriate to include TestNG count in this total, though I will ask @renfeiw to weigh in on this also.

ymanton:
Sorry, by count I meant classify/mark. Currently test targets whose test cases are skipped are marked as as failed. With this change they'll be marked as passing.

renfeiw:
Just wondering why do we want to keep a test target whose all test cases are skipped? Should we just remove this test or exclude this test on that platform?

ymanton:
I want to be able to decide to skip tests at runtime (based on whether or not the JDK being tested supports certain features, the environment, etc), and in a more granular way, rather than at test build time.

renfeiw:
Could you please give me an example that you need to skip the test cases in the runtime while it cannot be done in the playlist.xml? TKG is a general framework not only for testng, your change may have issue when non-testng tests fail with return code 2.

smlambert:
In my opinion it is important not to hide 'SKIPPED' status into 'PASSED' status, because PASSED implies the tests were run.  SKIPPED clearly shows the tests were not run.  The 2 states are distinct and should not be muddied together.

ymanton:
> Could you please give me an example that you need to skip the test cases in the runtime while it cannot be done in the playlist.xml? TKG is a general framework not only for testng, your change may have issue when non-testng tests fail with return code 2.

For my immediate concerns I can do what's needed in the playlist, but that seems less desirable because it will have to be repeated per playlist test, and the commands in most of the tests in the playlists are already hard to read. This is a much tidier solution that doesn't encourage duplication.

However if we want TKG to not be aware of TestNG's exit codes then I guess it will have to be.

> In my opinion it is important not to hide 'SKIPPED' status into 'PASSED' status, because PASSED implies the tests were run. SKIPPED clearly shows the tests were not run. The 2 states are distinct and should not be muddied together.

Agreed. I'm suggesting turning FAILED with exit code 2 to PASSED (or some other non-failure status) because TestNG returns exit code 2 when all test cases are skipped and we consider all non-zero exit codes as failing.

renfeiw:
As mentioned in the previous comment, this change may cause issue for non-testng tests. 

>For my immediate concerns I can do what's needed in the playlist, but that seems less desirable because it will have to be repeated per playlist test

Could you give us a specific example of the need for repeating the change for all playlist tests? It is unclear to me why your change will cause all playlist tests to skip.

ymanton:
> Could you give us a specific example of the need for repeating the change for all playlist tests? It is unclear to me why your change will cause all playlist tests to skip.

Sorry, I think you've misunderstood me. I wasn't suggesting that our existing playlists had to change. I'm suggesting that we can either make this kind of change in one place (like `settings.mk`) or in potentially many places (like `playlist.xml`) and that making it in one place is far better than making it in many places.

However, if we have to constrain ourselves to only allowing exit code 0 to signal success then I guess that's a limitation we'll have to accept. I can change the playlist to handle the exit code before invoking `TEST_STATUS` instead, or do something else to work around it.

ymanton:
I've worked around it in the playlist and no longer require this change.

PR #42, opened 2020-02-21
Add support for RISC-V

Willsparker:
Signed-off-by: William Parker william.parker1@ibm.com

### DIFF
--- a/src/org/openj9/envInfo/JavaInfo.java
+++ b/src/org/openj9/envInfo/JavaInfo.java
@@ -58,6 +58,8 @@ public class JavaInfo {
             spec += "_arm";
         } else if (osArch.contains("sparcv9")) {
             spec += "_sparcv9";
+        } else if (osArch.contains("riscv")) {
+            spec += "_riscv";
         } else {
             System.out.println("Cannot determine System.getProperty('os.arch')=" + osArch + "\n");
             return null;


src/org/openj9/envInfo/JavaInfo.java
```diff
@@ -102,7 +104,7 @@ public String getJDKImpl() {
         impl = impl.toLowerCase();
         if (impl.contains("ibm")) {
             return "ibm";
-        } else if (impl.contains("openj9")) {
+        } else if (impl.contains("openj9") || impl.contains("OpenJ9")) {
```
llxia: `impl` is set to toLowerCase() at line 104. I do not think this change is needed.

src/org/openj9/envInfo/JavaInfo.java
```diff
@@ -58,6 +58,8 @@ public String getSPEC() {
             spec += "_arm";
         } else if (osArch.contains("sparcv9")) {
             spec += "_sparcv9";
+	} else if (osArch.contains("riscv")) {
+	    spec += "_riscv";
```
llxia: Please fix the indentation

llxia:
Could you sign off your commit? And do we have any existing issues that this PR can link to? Thanks

src/org/openj9/envInfo/JavaInfo.java
Willsparker: I was seeing this message before I put the change in:
```
Cannot determine System.getProperty('os.arch')=riscv

System.getProperty('java.vm.name')=Eclipse OpenJ9 VM
```

src/org/openj9/envInfo/JavaInfo.java
Willsparker: Fixed

Willsparker:
ref: https://github.com/eclipse/openj9/issues/7421

src/org/openj9/envInfo/JavaInfo.java
llxia: Please revert the above change and try again.

Willsparker: Ah, sorry, I've just seen it prints it before it's set to `toLowerCase()` and that'll be what I'm seeing.

PR #49, opened 2020-04-02
Fix the issue that only one dump archived on zos

### LINKED ISSUE #48: core files for failures on z/OS
We did some work to collect core files on z/OS. I'm thinking it only works for one core file.

Looking at the artifact from the last zos nightly build, it only contains one core for the cmdLineTester_SCCommandLineOptionTests_1 test, but should contain 7 cores.
job/Test_openjdk11_j9_sanity.functional_s390x_zos_Nightly/234/

@renfeiw said:
Here is the pr to capture core from test: https://github.com/AdoptOpenJDK/TKG/pull/6 Maybe the script moveDmp.pl doesn't capture all cores, fyi @Lan-Xia.

renfeiw:
- loop through regex matches if multiple dumps found

Fixes #48

### DIFF
--- a/scripts/moveDmp.pl
+++ b/scripts/moveDmp.pl
@@ -133,19 +133,16 @@ sub moveTDUMPS {
 	}
 	logMsg("Attempting to move the TDUMPS to '$moveLocation', using the log to identify the TDUMP name to be moved");
 	if ($file) {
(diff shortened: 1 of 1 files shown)


llxia:
Could you provide Grinder job number for the internal testing? Thanks

renfeiw:
it's Grinder_TKG/244

PR #52, opened 2020-04-08
z/OS needs jdk/lib in ADD_JVM_LIB_DIR_TO_LIBPATH

### LINKED ISSUE #51: z/OS OpenJ9 ADD_JVM_LIB_DIR_TO_LIBPATH needs jdk/lib on LIBPATH
job/Test_openjdk11_j9_sanity.functional_s390x_zos_Nightly
jsigjnitest_0
shrtest_zos_0
thrstatetest_0
`Fatal error: JVM interface version mismatch: expecting 4, got 6.`

The problem is the jdk/lib directory isn't on the LIBPATH. z/OS uses an additional library libwrappers.so. When jdk/lib isn't on the LIBPATH the tests would fail like
```
EDC5205S DLL module not found. (errno2=0xC40B0025)
iconv_init: libwrappers.so not found
```
however there are JDK8 directories on the pre-existing LIBPATH, such as `/home/jenkins/openj9_resources/J8.0_64/lib/s390x`. The test finds the JDK8 library instead of JDK11 (due to the missing lib dir) and this results in the `version mismatch`.

pshipton:
In order to find libwrappers.so

Closes #51

Testing
- shrtest job/Grinder_zos/9 - passed
- sanity.functional job/Grinder_zos/10 - no test summary was created

### DIFF
--- a/openj9Settings.mk
+++ b/openj9Settings.mk
@@ -95,6 +95,7 @@ else
 		else
 			VM_SUBDIR_PATH=$(TEST_JRE_LIB_DIR)$(D)$(ARCH_DIR)$(D)$(VM_SUBDIR)
 			J9VM_PATH=$(TEST_JRE_LIB_DIR)$(D)$(ARCH_DIR)$(D)j9vm
+			LIB_PATH=$(TEST_JRE_LIB_DIR)$(D)$(ARCH_DIR)
 		endif
 	else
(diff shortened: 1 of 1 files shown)


llxia:
FYI @renfeiw Grinder_zos/10 does not have `TEST TARGETS SUMMARY`. Maybe something we need to look into.

PR #55, opened 2020-04-17
Generate parallel list locally

renfeiw:
- running mode parallelList, it will split tests and generate parallelList.mk
- we can run testList_<index> target in parallelList.mk to invoke splitted list
- add new mode: parallelList and additional options
- rename MkTreeGen to DirectoryWalker and modify it to support both modes
- restructure MkGen and separate PlaylistInfo from it
- add TestDivider to handle parallelList generation
- modify makefiles and build files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,9 +20,8 @@ test_output_*
 
 # misc auto generated files
 autoGen.mk
-dependencies.mk
 utils.mk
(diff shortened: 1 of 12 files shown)


renfeiw:
tested Internally at: Grinder_TKG/430,431,432,434,439.

PR #58, opened 2020-04-21
make compileTools a separate target

llxia:
For parallel runs, genParallelList only needs to call compileTools

Issue: https://github.com/AdoptOpenJDK/openjdk-tests/issues/1563

Signed-off-by: lanxia <lan_xia@ca.ibm.com>

### DIFF
--- a/makefile
+++ b/makefile
@@ -53,10 +53,16 @@ endif
 endif
 
 #######################################
-# compile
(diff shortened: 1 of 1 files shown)


llxia:
Grinder_TKG/454/
Grinder_TKG/455/

PR #70, opened 2020-05-15
Verify test duration queried from TRSS

renfeiw:
- print out invalid test duration and use default value for such case


renfeiw:
Grinder_TKG/518

PR #83, opened 2020-06-15
Add RISC-V support in TKG

ChengJin01:
The change is used to set up the test environment
on RISC-V.

Signed-off-by: Cheng Jin <jincheng@ca.ibm.com>


ChengJin01:
The main changes in ottawa.csv:
(1) only enable NoOptions for RISC-V (all other modes are basically disabled)
(2) add mode195 to ensure it is only disabled on RISC-V.

Note: the value of SPEC uses riscv-64 instead of riscv64.
If we agree to change to riscv64, then https://github.com/AdoptOpenJDK/TKG/blob/master/src/org/openj9/envInfo/JavaInfo.java needs to be updated to support this.

public String getSPEC() {
 ...
        String model = System.getProperty("sun.arch.data.model");
        System.out.println("System.getProperty('sun.arch.data.model')=" + model + "\n");
        if (Integer.parseInt(model.trim())==64) {
            spec += "-64"; <----- to be updated for riscv64
            spec = spec.replace("arm-64", "aarch64");
        }

Reviewer: @llxia

FYI: @ShelleyLambert , @sxa

ChengJin01:
As confirmed by @smlambert, the SPEC for `RISC-V` has changed to `riscv64` to align with the naming of buildspec in OpenJ9.

PR #103, opened 2020-08-24
Simplify parallel list generation

renfeiw:
- merge local and TRSS solution


renfeiw:
tested locally and in the internal grinders: Grinder_TKG/743, Grinder_TKG/753

PR #110, opened 2020-11-05
moveDmp: stop using -n on tail as it is not required and breaks on Solaris

sxa:
I don't believe any platforms require the `-n` to be specified on `tail` so this shouldn't impact anything else. I need this change in before I can let https://github.com/AdoptOpenJDK/openjdk-tests/pull/2039 go in (that PR currently hard codes the TKG repo to this branch, so once this PR is merged I can remove that from the other one)


smlambert (approved):
LGTM

PR #113, opened 2020-11-12
Move functional tests related 3rd party libs into functional level

OscarQQ:
related: https://github.com/AdoptOpenJDK/TKG/issues/63
Signed-off-by: OscarQQ <Yixin.Qian@ibm.com>

### DIFF
--- a/makefile
+++ b/makefile
@@ -34,14 +34,6 @@ ifndef TEST_ROOT
 endif
 
 #######################################
-# download all dependent jars
-#######################################
-getdependency:
-	perl scripts$(D)getDependencies.pl -path '$(TEST_ROOT)$(D)TKG$(D)lib' -task default
-
-.PHONY: getdependency
-
-#######################################
 # run test
 #######################################
 _TESTTARGET = $(firstword $(MAKECMDGOALS))
@@ -81,9 +73,9 @@ endif
 # compile tools
 #######################################
 include moveDmp.mk
-COMPILE_TOOLS_CMD=ant -f .$(D)scripts$(D)build_tools.xml -DTEST_JDK_HOME=$(TEST_JDK_HOME)
+COMPILE_TOOLS_CMD=ant -f .$(D)scripts$(D)build_tools.xml -DTEST_JDK_HOME=$(TEST_JDK_HOME) -DTEST_ROOT=$(TEST_ROOT)
 
-compileTools: getdependency
+compileTools: 
 	$(RM) -r $(COMPILATION_OUTPUT); \
 	$(MKTREE) $(COMPILATION_OUTPUT); \
 	($(COMPILE_TOOLS_CMD) 2>&1; echo $$? ) | tee $(Q)$(COMPILATION_LOG)$(Q); \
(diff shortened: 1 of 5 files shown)


scripts/getDependencies.xml
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+-->
+
+<project name="getDependency" default="build" basedir=".">
+	<taskdef resource="net/sf/antcontrib/antlib.xml" />
+	<description>
+		Get Third Party Dependencies
+	</description>
+
+	<!-- set properties for this build -->
```
renfeiw, line 23: set default LIB property to all

makefile
```diff
@@ -36,10 +36,10 @@ endif
 #######################################
 # download all dependent jars
 #######################################
-getdependency:
-	perl scripts$(D)getDependencies.pl -path '$(TEST_ROOT)$(D)TKG$(D)lib' -task default
+# getdependency:
```
renfeiw, line 39: don't forget to remove the comment section above

scripts/getDependencies.pl
```diff
@@ -26,9 +26,11 @@
 my $path;
 # define task
 my $task = "default";
+my $dependencyList;
```
renfeiw, line 29: You probably want to initialize it to "". Otherwise the code might break when there's no dependencyList option.

llxia: Could we default to all? Only change if `dependencyList` is set by users. In this way, all the getDependency jobs do not need to update. i.e., https://ci.eclipse.org/openj9/job/test.getDependency/

scripts/build_tools.xml
```diff
@@ -27,7 +27,14 @@
 		<mkdir dir="${build}" />
 	</target>
 
-	<target name="compile" depends="init" description="Using java ${JDK_VERSION} to compile the source  ">
+	<target name="getDependency" depends="init" description="get third party dependency">
+		<subant target="build">
+			<property name="LIB" value="json_simple"/>
+			<fileset dir="./" includes="getDependencies.xml" />
+		</subant>
+	</target>
```
llxia, line 35: If we use `<import file="./getDependencies.xml"/>`, I think we can avoid defining the above target.

scripts/getDependencies.xml
```diff
@@ -0,0 +1,31 @@
...
+	<property name="LIB" value="all" />
+
+	<target name="build">
+		<exec executable="perl">
+			<arg line="getDependencies.pl -path ${TEST_ROOT}/TKG/lib -task default -dependencyList ${LIB}"/>
+		</exec>
+	</target>
```
llxia, line 30: - rename `build ` to something meaningful. For example, getDependentLibs

scripts/build_tools.xml
```diff
@@ -27,7 +27,14 @@
 		<mkdir dir="${build}" />
 	</target>
 
-	<target name="compile" depends="init" description="Using java ${JDK_VERSION} to compile the source  ">
+	<target name="getDependency" depends="init" description="get third party dependency">
+		<subant target="build">
+			<property name="LIB" value="json_simple"/>
```
OscarQQ, line 32: how could we define LIB if we are using import?

renfeiw (approved):
LGTM

PR #114, opened 2020-11-12
Update testBot disable function

renfeiw:
- use an xml parser to replace the regex match
- add variation disable
- add a format function


llxia (approved):
LGTM

PR #118, opened 2020-11-24
Support disabling impl and java version

renfeiw:
- add impl and version elements
- user can write a combination of impl, version and variation in disable block
- error msg on specifying one element more than once in one disable block


renfeiw:
tested locally and internally: TKG_Grinder/25, 26, 29

llxia (approved):
LGTM

PR #119, opened 2020-11-24
Fix generate parallel list for disabled targets

renfeiw:


renfeiw:
tested locally and internally: TKG_Grinder/930, 931

PR #125, opened 2020-12-03
Support auto disable impl and version

renfeiw:
- update testbot code to support impl and ver disable


llxia (approved):
LGTM

PR #139, opened 2021-01-12
Generate TAP file if there are no tests for the given target

andrew-m-leonard:
Fixes https://github.com/AdoptOpenJDK/TKG/issues/138
If no test targets match the given target, then a TAP file is still produced.

Signed-off-by: Andrew Leonard <anleonar@redhat.com>

### DIFF
--- a/scripts/resultsSum.pl
+++ b/scripts/resultsSum.pl
@@ -30,6 +30,7 @@ my $jdkImpl = "";
 my $buildList = "";
 my $spec = "";
 my $customTarget = "";
+my $totalCount = 0;
 
 for (my $i = 0; $i < scalar(@ARGV); $i++) {
 	my $arg = $ARGV[$i];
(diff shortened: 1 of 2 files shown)


andrew-m-leonard:
Successful netty: https://ci.adoptopenjdk.net/view/Failing%20Test%20Jobs/job/Test_openjdk11_hs_sanity.external_x86-64_linux_netty/2/console

karianna (approved):
LGTM

andrew-m-leonard:
@smlambert Do we need another PR to generate a different TAP in this scenario?

llxia:
It was designed on purpose if there is no test match, no TAP will be generated and the Jenkins build will fail. This PR changed the default behavior. FYI @renfeiw

renfeiw:
Agreed with Lan's comment, it was designed to fail the build if no test found. For example, if someone uses a non existed target, the build needs to fail. BTW, I don't think you need to pass in totalCount, you can simply change this `($numOfTotal > 0)` to `($numOfTotal >=0)`.

smlambert:
I think the right approach is to revert this PR.  Andrew's PR https://github.com/AdoptOpenJDK/openjdk-tests/pull/2165 addressed the immediate [issue](https://github.com/AdoptOpenJDK/TKG/issues/138) this intended to fix.  

Andrew and I spoke on this, but it was my bad to not have closed this PR before it got inadvertently merged.

smlambert:
And to add context on this, we originally created TAP and it would show no tests were run.  While I feel this is a perfectly acceptable scenario, we received a lot of feedback from development team that they preferred that we fail if someone tries to run a make target that contains no tests.

llxia:
Thanks @smlambert . I will revert this PR as AdoptOpenJDK/openjdk-tests#2165 addressed the original issue.

PR #148, opened 2021-01-29
Update machineInfo detection

renfeiw:
- encapsulate detected info into Class
- add assertion for some of the prerequisites

Signed-off-by: renfeiw renfeiw@ca.ibm.com

### DIFF
--- a/src/org/openj9/envInfo/Info.java
+++ b/src/org/openj9/envInfo/Info.java
@@ -1,0 +1,29 @@
+/*******************************************************************************
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*      https://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*******************************************************************************/
+
+package org.openj9.envInfo;
+
+public class Info {
+	public String name;
+	public String[] cmds;
+	public String output;
+	public String req;
+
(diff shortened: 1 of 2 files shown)


renfeiw:
Grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_TKG/163, https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder_TKG/164
internal grinder: Grinder_TKG/1144

llxia (approved):
LGTM

smlambert (approved):
LGTM

smlambert:
I noticed while looking at the Grinders you launched that https://github.com/AdoptOpenJDK/TKG/blob/master/src/org/openj9/envInfo/MachineInfo.java#L35 does not work for osx, from a [quick query](https://stackoverflow.com/questions/1715580/how-to-discover-number-of-logical-cores-on-mac-os-x), looks like you can use `sysctl -n hw.ncpu`.  Can be addressed in a separate PR (since its previous code, and does not affect this PR).

From Grinder_TKG/163:

```
22:14:52  cpuCores : 
22:14:52  	cat: /proc/cpuinfo: No such file or directory
22:14:52  	       0
```

renfeiw:
> I noticed while looking at the Grinders you launched that https://github.com/AdoptOpenJDK/TKG/blob/master/src/org/openj9/envInfo/MachineInfo.java#L35 does not work for osx, from a [quick query](https://stackoverflow.com/questions/1715580/how-to-discover-number-of-logical-cores-on-mac-os-x), looks like you can use `sysctl -n hw.ncpu`. Can be addressed in a separate PR (since its previous code, and does not affect this PR).
> 
> From Grinder_TKG/163:
> 
> ```
> 22:14:52  cpuCores : 
> 22:14:52  	cat: /proc/cpuinfo: No such file or directory
> 22:14:52  	       0
> ```

Thanks for pointing out. I will address this in a separate issue. Looks like it might be helpful to have a way to specify different commands for different platforms.

PR #157, opened 2021-03-11
Ensure Q is defined in moveDmp.mk to prevent unquoted test on Solaris

sxa:
It may be the case that this should have already included the definition via [settings.mk](https://github.com/AdoptOpenJDK/TKG/blob/cca8f28c9c557037d8a04f7d097250fc53226d0d/settings.mk#L38) or https://github.com/AdoptOpenJDK/TKG/blob/cca8f28c9c557037d8a04f7d097250fc53226d0d/makeGen.mk#L28 but either way setting it in the file works. If there is a preferred solution to obtain it via the inclusion of one of the other files I'm happy to do that instead.

(As an aside this would also work if you ran with `make SHELL=bash compile` but for portability that's not the best option here unless things get problematic further down the line)


sxa:
For historic reference this was raisd as a result of [Grinder 7578](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/7578/console) failing with this error:
```
14:28:20  if [ -z $(tail -1 /export/home/jenkins/workspace/Grinder/openjdk-tests/TKG/../TKG/output_compilation/compilation.log | grep 0) ]; then perl scripts/moveDmp.pl --compileLogPath=/export/home/jenkins/workspace/Grinder/openjdk-tests/TKG/../TKG/output_compilation/compilation.log --testRoot=/export/home/jenkins/workspace/Grinder/openjdk-tests/TKG/..; false; else rm -f -r /export/home/jenkins/workspace/Grinder/openjdk-tests/TKG/../TKG/output_compilation; fi
14:28:20  /bin/sh: syntax error at line 4: `(' unexpected
14:28:20  gmake: *** [makefile:75: compileTools] Error 2
```

smlambert (commented):
Q is set here: https://github.com/AdoptOpenJDK/TKG/blob/master/settings.mk#L38 so would like to understand if/why we have a case were moveDmp is used before settings, and if there is a preferred way to handle it besides defining it twice

renfeiw:
moveDmp.mk doesn't include settings.mk at this moment. Since only Q is needed, I think it's not preferred to include setting.mk. So we can have Q added in moveDmp.mk like in this change. I think a better way is to create another small mk file which contains Q, D, etc. and we can include it wherever we need. But this can be a future improvement.

renfeiw (approved):

smlambert (approved):

PR #170, opened 2021-03-30
Add Run AQA (Comment-Triggered PR Builds) workflow

Icohedron:
As per AdoptOpenJDK/openjdk-tests#2399, adds the Run AQA (Comment-Triggered PR Builds) workflow to the TKG repo.
When ran on a PR in this (TKG) repo, the [AdoptOpenJDK/run-aqa@v1](https://github.com/AdoptOpenJDK/run-aqa) GitHub action is ran with the `tkg_Repo` argument set to the head of the PR.

### DIFF
--- a/.github/workflows/runAqaArgParse.py
+++ b/.github/workflows/runAqaArgParse.py
@@ -1,0 +1,78 @@
+import argparse
+import json
+import sys
+
+def map_platforms(platforms):
+    """ Takes in a list of platforms and translates Grinder platorms to corresponding GitHub-hosted runners.
+        This function both modifies and returns the 'platforms' argument.
+    """
+    
+    platform_map = {
+        'x86-64_windows': 'windows-latest',
+        'x86-64_mac': 'macos-latest',
+        'x86-64_linux': 'ubuntu-latest'
+    }
+    
+    for i, platform in enumerate(platforms):
+        if platform in platform_map:
+            platforms[i] = platform_map[platform]
+    
+    return platforms
+
+def underscore_targets(targets):
+    """ Takes in a list of targets and prefixes them with an underscore if they do not have one.
+    """
+    result = []
+    for target in targets:
+        t = target
+        if target[0] != '_':
+            t = '_' + t
+        result.append(t)
+    return result
+
+def main():
+
+    # The keyword for this command.
+    keyword = 'run aqa'
+    keywords = keyword.split()
+
+    # We assume that the first argument is/are the keyword(s).
+    # e.g. sys.argv == ['action_argparse.py', 'run', 'aqa', ...]
+    raw_args = sys.argv[1 + len(keywords):]
+
+    # Strip leading and trailing whitespace. Remove empty arguments that may result after stripping.
+    raw_args = list(filter(lambda empty: empty, map(lambda s: s.strip(), raw_args)))
+
+    parser = argparse.ArgumentParser(prog=keyword, add_help=False)
+    # Improvement: Automatically resolve the valid choices for each argument populate them below, rather than hard-coding choices.
+    parser.add_argument('--sdk_resource', default=['nightly'], choices=['nightly', 'releases', 'customized'], nargs='+')
+    parser.add_argument('--customized_sdk_url', default=['None'], nargs='+')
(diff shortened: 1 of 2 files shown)


lumpfish:
I've never written a github action, so perhaps not the best person to review this - but how can I invoke the action to test that it works?

smlambert:
Not sure why @karianna has assigned @lumpfish and @Mesbah-Alam to review this (perhaps thought this was the STF repo).  I will remove you from reviewers list.

This is part of the CanOSP work, and a port of what is already delivered to the openjdk-tests repo (and documented here: https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/doc/RunAqa.md)

@Icohedron has run a test in his branch, https://github.com/Icohedron/TKG/pull/1

karianna:
> Not sure why @karianna has assigned @lumpfish and @Mesbah-Alam to review this (perhaps thought this was the STF repo). I will remove you from reviewers list.
> 
> This is part of the CanOSP work, and a port of what is already delivered to the openjdk-tests repo (and documented here: https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/doc/RunAqa.md)
> 
> @Icohedron has run a test in his branch, [Icohedron#1](https://github.com/Icohedron/TKG/pull/1)

Yep in my tired state I read STF :-/

.github/workflows/runAqaArgParse.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Strip leading and trailing whitespace. Remove empty arguments that may result after stripping.
+    raw_args = list(filter(lambda empty: empty, map(lambda s: s.strip(), raw_args)))
+
+    parser = argparse.ArgumentParser(prog=keyword, add_help=False)
+    # Improvement: Automatically resolve the valid choices for each argument populate them below, rather than hard-coding choices.
+    parser.add_argument('--sdk_resource', default=['nightly'], choices=['nightly', 'releases', 'customized'], nargs='+')
+    parser.add_argument('--customized_sdk_url', default=['None'], nargs='+')
+    parser.add_argument('--archive_extension', default=['.tar'], choices=['.zip', '.tar', '.7z'], nargs='+')
+    parser.add_argument('--build_list', default=['openjdk'], choices=['openjdk', 'functional', 'system', 'perf', 'external'], nargs='+')
```
smlambert, line 51: I saw this in openjdk-tests repo, but not tried yet to find out what happens when someone tries to build external?  Is docker installed on github runners and this works smoothly?

sophia-guo: I've tried with jenkin_tests and camel tests and hit the same issue:
https://github.com/sophia-guo/runaqaTest/runs/2249338597?check_suite_focus=true
https://github.com/sophia-guo/runaqaTest/runs/2249127830?check_suite_focus=true

Look docker is fine. There are some hidden issues related with environment setup? Both jobs can successfully create docker images but some how `unable to recognize DOCKER_OS` hence docker run is trying to delete default images and find the image is not there. I will take a look.

```
Unable to recognize DOCKER_OS from DOCKERIMAGE_TAG = --version!
docker run  --rm adoptopenjdk-derby-test:8-jdk-ubuntu-hotspot-full 
Unable to find image 'adoptopenjdk-derby-test:8-jdk-ubuntu-hotspot-full' locally

```

renfeiw (approved):
The porting change looks good to me. For future improvement, I think it may be better to keep runAqaArgParse.py in a common place, for example under TKG/scripts/testBot, so we don't need to duplicate it in multiple repos.

Icohedron:
> The porting change looks good to me. For future improvement, I think it may be better to keep runAqaArgParse.py in a common place, for example under TKG/scripts/testBot, so we don't need to duplicate it in multiple repos.

The defaults and/or available parameters may differ for each repo. 
For this TKG repo version of the workflow, the `tkg_repo` argument is not available because the TKG repo will always be the head of the PR. In the openjdk-test repo version of this workflow, a custom `tkg_repo` is able to be specified.
Hence, the `runAqaArgParse.py` is slightly different between the two versions of the workflow for the two repos.

smlambert (approved):
Indeed, in future if we put the script in a common place, it could be refactored to read defaults from a config file for each repo its contained within, but certainly can come later.

Thanks @Icohedron !

PR #173, opened 2021-04-07
Add help command for a link to documentation in the Run AQA workflow

Icohedron:
Add a `--help` / `-h` argument to the Run AQA workflow (#170) which links to its documentation: 
https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/doc/RunAqa.md

image

This PR implements the same functionality of AdoptOpenJDK/openjdk-tests#2457

**Note for testing**: This PR changes `runAqaArgParse.py` but the workflow will continue to use the unchanged `runAqaArgParse.py` from `adoptium:master`. Therefore, for testing, [line 29 of `runAqa.yml`](https://github.com/adoptium/TKG/blob/master/.github/workflows/runAqa.yml#L29) should be changed to match your/my personal repo which has the updated `runAqaArgParse.py`.

### DIFF
--- a/scripts/testBot/runAqaArgParse.py
+++ b/scripts/testBot/runAqaArgParse.py
@@ -79,6 +79,8 @@ Click the above link to view the documentation for the Run AQA GitHub Workflow''
         sys.stderr.write(help_msg)
         print('::error ::Help command invoked')
         exit(2)
+    # Remove help flag from the args because it is not a parameter for the job matrix.
+    del args['help']
 
     # Map grinder platform names to github runner names
     args['platform'] = map_platforms(args['platform'])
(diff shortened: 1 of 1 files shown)


llxia (approved):
LGTM

smlambert (approved):
thanks @Icohedron !

scripts/testBot/runAqaArgParse.py
```diff
@@ -65,6 +66,22 @@ def main():
...
+        help_msg = '''Run AQA GitHub Action Documentation
+`\\`\\`
+https://github.com/AdoptOpenJDK/openjdk-tests/blob/master/doc/RunAqa.md
+\\`\\`\\`
+Click the above link to view the documentation for the Run AQA GitHub Workflow'''
+        sys.stderr.write(help_msg)
+        print('::error ::Help command invoked')
+        exit(2)
+    # Remove help flag from the args because it is not a parameter for the job matrix.
+    del args['help']
```
renfeiw, line 83: Just wonder is it necessary to del help arg? If it's defined, the program is exit now. If not, do we need to del it?

Icohedron: If `--help` is not specified then it will be stored a value of False.

Since the args get converted to JSON for the runBuild job matrix to parse, it will read `help: false` and throw an error because `help` is not an array of values.

renfeiw (approved):

PR #174, opened 2021-04-07
Fix default value for openjdk_testrepo in Run AQA workflow script

Icohedron:
The previous default value for `--openjdk_testrepo` did not include a branch and therefore was invalid, which caused builds using the default openjdk_testrepo to fail.

**Note for testing**: This PR changes `runAqaArgParse.py` but the workflow will continue to use the unchanged `runAqaArgParse.py` from `adoptium:master`. Therefore, for testing, [line 29 of `runAqa.yml`](https://github.com/adoptium/TKG/blob/master/.github/workflows/runAqa.yml#L29) should be changed to match your/my personal repo which has the updated `runAqaArgParse.py`.


llxia (approved):
LGTM

renfeiw (approved):
LGTM

PR #175, opened 2021-04-09
Run AQA Workflow: Implement build-jdk sdk_resource, Remove redundant checkouts

Icohedron:
This PR builds on top of #170: Run AQA (Comment-Triggered PR Build) workflow

- Change the `Echo parameters` step to simply echo the entire JSON of build parameters instead of each individual value in addition to it.

- Reduce redundant checkouts by using the fact that the current repo is the TKG repo. This also removes the need to modify the TKG checkout action to test changes to `runAqaArgParse.py`.

- Implement a new sdk_resource `build-jdk` which uses the [AdoptOpenJDK/build-jdk](https://github.com/AdoptOpenJDK/build-jdk/) action to build a JDK using a provided openjdk-build repo.

This functionality was requested in issue https://github.com/AdoptOpenJDK/openjdk-tests/issues/2296 but was blocked due to issue https://github.com/AdoptOpenJDK/build-jdk/issues/24 which is still unresolved, but has a (unintended?) workaround that works due to the `usePRRef` input not causing the action to pull from the PR ref but instead have the action use whatever existing `openjdk-build` directory is in the workspace (if any).

The same workaround does not work for the [eclipse-openj9/build-openj9](https://github.com/eclipse-openj9/build-openj9) action because its similarly-named `usePersonalRepo` input does checkout the PR ref and also does not accept a custom repo and branch input (https://github.com/eclipse-openj9/build-openj9/issues/18), so the action would need to be rewritten.

Example run: [734853617](https://github.com/Icohedron/TKG/actions/runs/734853617)
image

### DIFF
--- a/.github/workflows/runAqa.yml
+++ b/.github/workflows/runAqa.yml
@@ -17,22 +17,15 @@ jobs:
         echo ::set-output name=url::$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID
         echo ::set-output name=id::$GITHUB_RUN_ID
       id: workflow_run_info
-    # Checkout current repo to access the repo-specific config file `.github/workflows/runAqaConfig.json`
+    # Checkout current (TKG) repo to access the repo-specific config file `.github/workflows/runAqaConfig.json` and the shared script `scripts/testRepo/runAqaArgParse.py`
     - name: Checkout current repo
       uses: actions/checkout@v2
       with:
         path: 'main'
-    # Checkout the main TKG repo to access the shared script `scripts/testRepo/runAqaArgParse.py`
-    - name: Checkout main TKG repo
-      uses: actions/checkout@v2
-      with:
-        repository: 'adoptium/TKG.git'
-        ref: 'master'
-        path: 'TKG'
     - name: Parse parameters
       env:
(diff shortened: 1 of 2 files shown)


llxia (approved):
LGTM

llxia:
@renfeiw or @sophia-guo can either of you view this PR? Thanks

renfeiw (approved):
LGTM

sophia-guo:
> Implement a new sdk_resource build-jdk which uses the AdoptOpenJDK/build-jdk action to build a JDK using a provided openjdk-build repo.

The new added sdk_resource enabled to build a JDK using a provided openjdk-build repo. However the JDK isn't setup as TEST_JDK_HOME and used by run-aqa https://github.com/Icohedron/TKG/blob/d39cb9dab495526fd0eeb1ad6c8e9c2b2aaa81e4/.github/workflows/runAqa.yml#L151. The example of run-aqa use a upstream build jdk:

https://github.com/adoptium/run-aqa
```
    - name: AQA
      uses: AdoptOpenJDK/run-aqa@v1
      env:
         TEST_JDK_HOME: ${{ steps.buildOpenj9.outputs.BuildOpenJ9JDK }}
      with: 
        build_list: 'openjdk'
        target: '_jdk_math'
```

Issue #61, opened 2020-04-28
Remove TEST_RESULTSTORE and related logic


llxia:
We are using artifactory to store test outputs. I think the logic for `TEST_RESULTSTORE` can be removed.

https://github.com/AdoptOpenJDK/TKG/blob/9123d8f0f12600b34530e2ac2b0093b5d231b179/openj9Settings.mk#L137

SAY-droid427:
@llxia @smlambert  May I work on this issue?

llxia:
Close this issue for now. We will revisit it when the internal change is in.

PR #183, opened 2021-04-16
Removed TEST_RESULTSTORE logic

SAY-droid427:
As per my understanding of the issue, I have removed the [lines 135 - 151](https://github.com/adoptium/TKG/blob/master/openj9Settings.mk#L135-L151).

Fixes https://github.com/adoptium/TKG/issues/61


SAY-droid427:
@smlambert Please review the PR and let me know if anything else had to be done. Thanks!

smlambert (approved):
thanks @SAY-droid427 !

renfeiw:
Thanks @SAY-droid427. Discussed with @llxia, we need to further verify it will not break any internal testing first.

SAY-droid427:
Thanks for taking out the time for reviewing the PR!

karianna (approved):
LGTM

renfeiw:
Looks like we are still using this feature internally, so we cannot remove it at this moment. We need to port it to internal first if we want to remove it. fyi @llxia

llxia:
The original issue #61 was created at the time when the logic was in both internal and external version. Now, this logic is in external only, we cannot remove it. As @renfeiw mentioned, in order to remove it, we need to port the logic to internal first. Otherwise, vmfarm upload will break.

karianna:
@SAY-droid427 - merge conflict

SAY-droid427:
@karianna Resolved merge conflicts

karianna:
@llxia has that internal change landed?

llxia:
> @llxia has that internal change landed?

No, not yet

karianna:
Converting to draft so we don't merge by accident until it's ready.

PR #187, opened 2021-04-27
Removed unnecessary macros

### LINKED ISSUE #159: Remove unnecessary macros
Due to historical reasons, we have `D` and `Q` macros defined in makefile in TKG and used in test repos.

For example:
https://github.com/AdoptOpenJDK/TKG/blob/master/compile.mk#L17
https://github.com/AdoptOpenJDK/TKG/blob/master/makeGen.mk#L26
https://github.com/AdoptOpenJDK/TKG/blob/master/makeGen.mk#L28

I think we should verify and remove them from makefiles and test repos if they are unnecessary.

2001asjad:
Added a common `makefile` in the project repository for the common macros like D, P, Q etc are defined and they have been included in the rest of the makefiles using `include` .

Fixes #159


2001asjad:
Hello @smlambert @llxia @renfeiw , I've made a PR. Looking forward for your reviews so that we can make it better.
Thank You

scripts/testBot/runAqaArgParse.py
```diff
@@ -1,22 +1,26 @@
 import argparse
 import json
 import sys
+import os
+
+dir_list = next(os.walk('.'))[1]
+
```
llxia, line 7: Please do not include other changes in this PR. Thanks

.github/workflows/runAqa.yml
```diff
@@ -26,8 +26,8 @@ jobs:
     - name: Checkout main TKG repo
       uses: actions/checkout@v2
       with:
-        repository: 'adoptium/TKG.git'
-        ref: 'master'
+        repository: '2001asjad/TKG.git'
+        ref: 'asjad-issue-2482'
```
llxia, line 30: Please revert reference to personal repo/branch

2001asjad:
Oh, I am sorry @llxia , I was also working on another issue in this repo, let me fix this.
Are rest of the things fine, like the ones that are concerned with the isse.

PR #226, opened 2021-07-08
Add mixed riscv spec

renfeiw:


renfeiw:
grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1053/

llxia:
In original riscv specs, most modes were purposely turn off. But the basic modes (Mode150, 650) need to turn on in this PR. Otherwise, tests with these modes will not run. For example:
https://github.com/adoptium/aqa-tests/blob/master/openjdk/playlist.xml#L42

renfeiw:
> In original riscv specs, most modes were purposely turn off. But the basic modes (Mode150, 650) need to turn on in this PR. Otherwise, tests with these modes will not run. For example:
> https://github.com/adoptium/aqa-tests/blob/master/openjdk/playlist.xml#L42

Updated, thanks.

llxia (approved):
LGTM

smlambert:
Grinder looks good: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/1058/

smlambert (approved):
thanks @renfeiw !

PR #234, opened 2021-08-31
Clean build files before compilation

renfeiw:

### DIFF
--- a/makefile
+++ b/makefile
@@ -56,13 +56,13 @@ endif
 # compile test materials
 #######################################
 compile: envDetect
+	$(MAKE) -f clean.mk cleanBuild
 	$(MAKE) -f compile.mk compile
 
 #######################################
 # If AUTO_DETECT is turned on, compile and execute envDetector in build_envInfo.xml.
 #######################################
 envDetect: compileTools
-
 	${TEST_JDK_HOME}$(D)bin$(D)java -cp .$(D)bin$(D)TestKitGen.jar org.openj9.envInfo.EnvDetector
 
 #######################################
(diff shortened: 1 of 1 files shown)


renfeiw:
related PR: https://github.com/eclipse-openj9/openj9/pull/13419

smlambert (approved):
LGTM

llxia (approved):
LGTM

keithc-ca:
What is the motivation for this change?

Previously, one could say `make compile` to incrementally recompile test code, but now compiles start from scratch every time. This will negatively affect developer productivity.

Why not just add a `cleanBuild` target that can be used where necessary?

keithc-ca:
On a system I often use, this adds over two minutes to the build time (for no changes):
* before: `Total time: 4 minutes 13 seconds`
* after:  `Total time: 6 minutes 37 seconds`

renfeiw:
> What is the motivation for this change?
> 
> Previously, one could say `make compile` to incrementally recompile test code, but now compiles start from scratch every time. This will negatively affect developer productivity.
> 
> Why not just add a `cleanBuild` target that can be used where necessary?

This is the first step towards dynamic compilation. To be more specific, this is used to check if a folder is compiled by checking if the jar file exist. I don't think the previous version can incrementally compile the code. In the previous behaviour, there is no check, even if the jar file exist it is still compiled from scratch. The current behaviour is if a dependent jar file is compiled, we don't compile it. So there should be no more files compiled compared to the previous version. 

If you are doing a development locally, you can export the BUILD_LIST to limit the folders to be compiled. With the current version, you don't need to provide dependent folders in the BUILD_LIST. 
For example, if you want to run and compile forceLazySymbolResolution test, you just need to do 
```export BUILD_LIST=functional/cmdLineTests/forceLazySymbolResolution```, 
in the previous version one has to do
 ```export BUILD_LIST=functional/cmdline_options_tester, functional/cmdline_options_testresources,functional/VM_Test,functional/cmdLineTests/forceLazySymbolResolution```

keithc-ca:
Thanks for the explanation. I sure everyone is willing to put up with some short-term pain if in the end we have quicker edit/compile cycles.

PR #253, opened 2021-11-30
Allow other jenkins parameters to show in Grinder re-run console links

sxa:
This change will make it easier to avoid resubmitting jobs with incorrect parameters when using the Grinder links at the bottom of the console log output. There are two changes here:
1. Previously on 32-bit ARM runs, the Grinder link would set `PLATFORM` to `aarch32_linux`. This is incorrect and this PR adjusts that to be `arm_linux`
2. Some of the parameters use by jenkins, will now be added to the Grinder parameters if they are located in the environment (as they will be for jobs run through jenkins). These will allow the acquisition of the correct source materials used for the testing to take place. The following environment variables are the ones that will now be captured in the grinder re-run logs in the bottom of the console log: `SDK_RESOURCE`, `CUSTOMIZED_SDK_URL`, `CUSTOMIZED_SDK_URL_CREDENTIAL_ID`, `UPSTREAM_JOB_NAME`, `UPSTREAM_JOB_NUMBER` and the TCK-specific `JCK_GIT_REPO`.

### DIFF
--- a/resources/buildPlatformMap.properties
+++ b/resources/buildPlatformMap.properties
@@ -10,6 +10,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# NOTE: These mappings are used exclusively for the "Re-run in Grinder"
+#       links in the console log output
+
 aix_ppc-64_nocmprssptrs=ppc64_aix_xl
 aix_ppc-64_cmprssptrs=ppc64_aix_cmprssptrs
 aix_ppc-64=ppc64_aix
@@ -21,7 +24,7 @@ linux_390=s390_linux
 linux_aarch64_nocmprssptrs=aarch64_linux_xl
 linux_aarch64_cmprssptrs=aarch64_linux_cmprssptrs
 linux_aarch64=aarch64_linux
-linux_arm=aarch32_linux
+linux_arm=arm_linux
 linux_ppc-64_nocmprssptrs=ppc64_linux_xl
 linux_ppc-64_cmprssptrs=ppc64_linux_cmprssptrs
 linux_ppc-64_cmprssptrs_le=ppc64le_linux_cmprssptrs
--- a/resources/buildPlatformMap.properties
+++ b/resources/buildPlatformMap.properties
@@ -10,9 +10,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# NOTE: These mappings are used exclusively for the "Re-run in Grinder"
-#       links in the console log output
-
 aix_ppc-64_nocmprssptrs=ppc64_aix_xl
 aix_ppc-64_cmprssptrs=ppc64_aix_cmprssptrs
 aix_ppc-64=ppc64_aix
--- a/scripts/resultsSum.pl
+++ b/scripts/resultsSum.pl
@@ -250,8 +250,6 @@ sub resultReporter {
 		my $platformParam = "";
 		if (exists $spec2platform->{$spec}) {
 			$platformParam = "&PLATFORM=" . $spec2platform->{$spec};
-			# spec fileuses aarch32, but Grinder needs arm as the parameter
-			$platformParam =~ 's/aarch32_/arm_/';
 		}
 		my $customTargetParam = "";
 		if ($customTarget ne '') {
--- a/scripts/resultsSum.pl
+++ b/scripts/resultsSum.pl
@@ -263,6 +263,8 @@ sub resultReporter {
 		if ($jdkVendor ne '') {
 			$vendorParam = "&JDK_VENDOR=" . $jdkVendor;
 		}
+		# TODO: Revisit this and make it more generic and allow all potentially
+		#       applicaable Grinder job paramers to be made availablea
 		my $jenkinsParam = "";
 		if ( defined $ENV{'JCK_GIT_REPO'} ) {
 			$jenkinsParam .= "&JCK_GIT_REPO="        . $ENV{'JCK_GIT_REPO'};
--- a/scripts/resultsSum.pl
+++ b/scripts/resultsSum.pl
@@ -264,7 +264,7 @@ sub resultReporter {
 			$vendorParam = "&JDK_VENDOR=" . $jdkVendor;
 		}
 		# TODO: Revisit this and make it more generic and allow all potentially
-		#       applicaable Grinder job paramers to be made availablea
+		#       applicable Grinder job parameters to be made available
 		my $jenkinsParam = "";
 		if ( defined $ENV{'JCK_GIT_REPO'} ) {
 			$jenkinsParam .= "&JCK_GIT_REPO="        . $ENV{'JCK_GIT_REPO'};
--- a/scripts/resultsSum.pl
+++ b/scripts/resultsSum.pl
@@ -250,6 +250,8 @@ sub resultReporter {
 		my $platformParam = "";
 		if (exists $spec2platform->{$spec}) {
 			$platformParam = "&PLATFORM=" . $spec2platform->{$spec};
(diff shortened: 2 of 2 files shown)


sxa:
Fixes TC issue 130 which as discussed in the retrospective last week.

scripts/resultsSum.pl
```diff
@@ -231,6 +231,8 @@ sub resultReporter {
 		my $platformParam = "";
 		if (exists $spec2platform->{$spec}) {
 			$platformParam = "&PLATFORM=" . $spec2platform->{$spec};
+			# spec fileuses aarch32, but Grinder needs arm as the parameter
+			$platformParam =~ 's/aarch32_/arm_/';
```
renfeiw, line 235: If the aarch32 is replaced by arm, I believe we need to update the map file: https://github.com/adoptium/TKG/blob/master/resources/buildPlatformMap.properties#L24, instead of change it here. @llxia What do you think?

sxa: I wasn't expecting that we change it universally for now, just set the "correct" Grinder parameter as a tactical fix to it giving an invalid link. If we choose to do a larger change later that should probably be a separate PR

llxia: This is not the correct place to change it. The update should be in https://github.com/adoptium/TKG/blob/master/resources/buildPlatformMap.properties#L24 as @renfeiw mentioned above. It is one line change.

scripts/resultsSum.pl
```diff
@@ -244,11 +246,31 @@ sub resultReporter {
...
+			$jenkinsParam .= "&CUSTOMIZED_SDK_URL_CREDENTIAL_ID=" . $ENV{'CUSTOMIZED_SDK_URL_CREDENTIAL_ID'};
+		}
+		if ( defined $ENV{'UPSTREAM_JOB_NAME'} ) {
+			$jenkinsParam .= "&UPSTREAM_JOB_NAME="   . $ENV{'UPSTREAM_JOB_NAME'};
+		}
+		if ( defined $ENV{'UPSTREAM_JOB_NUMBER'} ) {
+			$jenkinsParam .= "&UPSTREAM_JOB_NUMBER=" . $ENV{'UPSTREAM_JOB_NUMBER'};
+		}
+		
+		my $rebuildLinkBase = "parambuild/?JDK_VERSION=$jdkVersion&JDK_IMPL=$jdkImpl$vendorParam$buildParam$platformParam$customTargetParam$jenkinsParam";
```
llxia, line 269: The above Jenkins parameters changes are not necessary.

For rerun in Jenkins, please use the `Rerun in Grinder` link on the Jenkins job. It captures all Jenkins parameters. For example, in the Grinder https://ci.adoptopenjdk.net/job/Grinder/2701/

image

The rerun link in TKG is added for local runs. It is an easy way for the local run users to validate/compare their change in the farm. TKG is designed to be CI agnostic and supports the minimal Jenkins parameters in the TKG layer.

scripts/resultsSum.pl
llxia: This is not the correct place to change it. The update should be in https://github.com/adoptium/TKG/blob/master/resources/buildPlatformMap.properties#L24 as @renfeiw mentioned above. It is one line change.

sxa: Will that not change other things? I am not looking to use this PR to change what the platform is called universally, only to fix a situation where an invalid parameter is appearing on re-run jobs. Will the `buildPlatformMap.properties` only affect that, or cause other (potentially breaking) changes if that was the only place it was changed?

scripts/resultsSum.pl
smlambert: Stewart, I've checked and most of these will not work as you expect as they are not set as environment variables.  The only one that is set is JCK_GIT_REPO and even if you do put it in the list, you will still need to add the credential to access it (which is not set as an environment variable).  

You really won't gain much by this section, so it's mainly adding code that will never be invoked.  

Best to use the link available from the Jenkins page and not this link that is intended to be used only for local runs (and subsequently needs many more additional parameters to be filled in by the user).

smlambert (approved):
LGTM

sxa:
Running a Grinder with the change at https://ci.adoptopenjdk.net/job/Grinder/2984/console

resources/buildPlatformMap.properties
```diff
@@ -10,6 +10,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# NOTE: These mappings are used exclusively for the "Re-run in Grinder"
+#       links in the console log output
+
```
renfeiw, line 15: You should delete this comment, as it's not only for rerun link (also used for exclusion etc). This file should match whatever is defined in https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/openjdk_tests

sxa: OK ... Shelley's earlier comment had said:

> buildPlatformMap.properties is used for rerun in Grinder link alone

but I can remove it.

smlambert: my misunderstanding... Now I need to know how are we passing other tests without this change and will have to look and understand it better, dangit! ;)

renfeiw: I guess no one is using arm_linux for exclude, so there's no break till now :)

renfeiw: I think it is not necessary to add this comment, it is also used by other things such as internal TKG and etc.

sxa:
> Running a Grinder with the change at https://ci.adoptopenjdk.net/job/Grinder/2994/console

Looks good: It came up with this Grinder link in [the console log](https://ci.adoptopenjdk.net/job/Grinder_testList_0/51/) to rebuild the failed `jdk_nio_1`:
- https://ci.adoptopenjdk.net/job/Grinder/parambuild/?JDK_VERSION=11&JDK_IMPL=hotspot&JDK_VENDOR=eclipse&BUILD_LIST=openjdk&PLATFORM=arm_linux&SDK_RESOURCE=upstream&UPSTREAM_JOB_NAME=build-scripts/jobs/jdk11u/jdk11u-linux-arm-hotspot&UPSTREAM_JOB_NUMBER=916&TARGET=jdk_nio_2

Previously this [extended.openjdk run](https://ci.adoptopenjdk.net/job/Test_openjdk11_hs_extended.openjdk_arm_linux_testList_2/21/console) gave this link without the additional parameters so I'm comfortable that this PR achives the desired goal:
- https://ci.adoptopenjdk.net/job/Grinder/parambuild/?JDK_VERSION=11&JDK_IMPL=hotspot&JDK_VENDOR=eclipse&BUILD_LIST=openjdk&PLATFORM=aarch32_linux&TARGET=jdk_nio_2

scripts/resultsSum.pl
sxa: Based on the tests in https://github.com/adoptium/TKG/pull/253#issuecomment-1010419210 it seems to be working as I wanted it to when run through jenkins :-)

scripts/resultsSum.pl
```diff
@@ -263,11 +263,33 @@ sub resultReporter {
 		if ($jdkVendor ne '') {
 			$vendorParam = "&JDK_VENDOR=" . $jdkVendor;
 		}
+		# TODO: Revisit this and make it more generic and allow all potentially
+		#       applicaable Grinder job paramers to be made availablea
```
smlambert, line 267:
```suggestion
		#       applicable Grinder job parameters to be made available
```

llxia:
Approve it for now. We will have further discussion of how to properly resolve the problem.

llxia (approved):

PR #257, opened 2022-01-11
Replace grep -p with standard UNIX awk operation

sxa:
Fixes https://github.com/adoptium/aqa-tests/issues/3208

Running https://ci.adoptopenjdk.net/job/Grinder/2982/console to check

Looks like the same line that caused a problem before has passed without problems:
```
14:33:52  /home/jenkins/workspace/Grinder/aqa-tests/TKG/scripts/getTestenvProperties.sh --repo_dir /home/jenkins/workspace/Grinder/aqa-tests --output_file /home/jenkins/workspace/Grinder/aqa-tests/testenv/testenv.properties --repo_name ADOPTOPENJDK
14:33:52  Check sha in /home/jenkins/workspace/Grinder/aqa-tests and store the info in /home/jenkins/workspace/Grinder/aqa-tests/testenv/testenv.properties
14:33:52  ADOPTOPENJDK_REPO=https://github.com/AdoptOpenJDK/openjdk-tests.git
14:33:52  ADOPTOPENJDK_BRANCH=11c721b16a5906f9e5b35e36b49b1d130d231a01
```

### DIFF
--- a/scripts/getTestenvProperties.sh
+++ b/scripts/getTestenvProperties.sh
@@ -91,7 +91,8 @@ getTestenvProperties() {
 		fi
 
 		# append the info into $OUTPUT_FILE
-		{	echo "${REPO_NAME}_REPO=$(git remote show origin -n | grep -Po '(?<=Fetch URL: ).*')";
+		{
+			echo "${REPO_NAME}_REPO=$(git remote show origin -n | awk '/Fetch URL:/{print $3}')";
 			echo "${branch}=${REPO_SHA}";
 		}  2>&1 | tee -a $OUTPUT_FILE
 	else


scripts/getTestenvProperties.sh
```diff
@@ -91,7 +91,8 @@ getTestenvProperties() {
 		fi
 
 		# append the info into $OUTPUT_FILE
-		{	echo "${REPO_NAME}_REPO=$(git remote show origin -n | grep -Po '(?<=Fetch URL: ).*')";
+		{
+			echo "${REPO_NAME}_REPO=$(git remote show origin -n | awk '/Fetch URL:/{print $3}')";
```
karianna, line 95: I'd add a comment as to why awk.

karianna (changes requested):

scripts/getTestenvProperties.sh
sxa: @karianna Hmmm I'm not sure that's necessary - why would you add a comment on why you're using any standard shell utility? It's just the right standard tool for the job. IMHO such a comment would only make sense when looking at this code change. Unclear on how a meaningful comment fo rthe casual reader who hadn't seen this PR could be written, but happy to take a suggestion.

smlambert: awk is the more prevalent utility, present in a known/stable form on all platforms we must test on, unlike as we just learned grep, which apparently is not available in a standard version/form across all platforms that we test on.  Agree that we do not need to explain the use of awk in a comment in the code, do not think it adds to the readability of the code.

smlambert (approved):
Grinders on the various platforms/versions look good. Thanks @sxa !

smlambert:
[fromPRtesting_testenv.properties.txt](https://github.com/adoptium/TKG/files/7848580/fromPRtesting_testenv.properties.txt)
testenv.properties file from an alpine-linux run now created properly

karianna (approved):

llxia:
re https://github.com/adoptium/TKG/pull/257#issuecomment-1011179507, the Grinder passed. Thanks @sxa

llxia (approved):
LGTM

PR #258, opened 2022-01-18
Add support for AArch64 macOS

knn-k:
This commit adds support for osx_aarch64_nocmprssptrs.

Signed-off-by: KONNO Kazuhiro <konno@jp.ibm.com>


karianna (approved):
LGTM

karianna:
@smlambert required for 17 release?

smlambert:
For AQAvit repositories, we have shifted to use a release branch (aqa-tests v0.8.0-release branch and updating the testenv.properties file in that branch, where TKG tag is v0.8.0) for the release.  In this way, we do not need to freeze the AQAvit repositories during a release, but changes to master will not get picked up.  

We do not need this change for the release, so, after some testing (and a 2nd review) it can go in without concern for affecting the release.

For informational purposes, I will add this note, if we did want it for the release, the steps we'd need would be to follow:
- merge it in 
- tag TKG with a v0.8.1 tag
- update the testenv.properties file to set the TKG_BRANCH=v0.8.1

knn-k:
OK, thank you.  I don't need this to be in the release branch.

knn-k:
I added the new `osx_aarch64_nocmprssptrs` column between `osx_x86-64` and `bsd_x86-64` in the csv file.
I ran the following two Grinder jobs to make sure that my change does not break other platforms, and they went successful:

- osx_x86-64: job/Grinder/20319/
- linux_aarch64: job/Grinder/20320/

I used `knn-k:aarch64macos1` as `TKG_OWNER_BRANCH` in these Grinder jobs.
I was not able to find existing jobs for `bsd_x86-64` and `sunos_*-64`.
Compressed refs and mixed refs will be added to AArch64 macOS later.

renfeiw (approved):

PR #259, opened 2022-01-21
Remove runAqaConfig.json

nadeensami:
Allow custom repos to be set by the users as command arguments rather than in runAqaConfig.json.

Fixes: adoptium/aqa-tests#3163

Signed-off-by: Nadeen Mohamed <nadeen@ualberta.ca>

### DIFF
--- a/.github/workflows/runAqaConfig.json
+++ b/.github/workflows/runAqaConfig.json
@@ -1,6 +1,3 @@
 {
-    "workflow_repo": "adoptium/TKG",
-    "custom_openjdk_testrepo": true,
-    "custom_openj9_repo": true,
-    "custom_tkg_repo": false
+	"workflow_repo": "adoptium/TKG"
 }
--- a/scripts/testBot/runAqaArgParse.py
+++ b/scripts/testBot/runAqaArgParse.py
@@ -55,18 +55,15 @@ def main():
     parser.add_argument('--platform', default=['x86-64_linux'], nargs='+')
     parser.add_argument('--jdk_version', default=['8'], nargs='+')
     parser.add_argument('--jdk_impl', default=['openj9'], choices=['hotspot', 'openj9'], nargs='+')
-
+    parser.add_argument('--openjdk_testrepo', default=['adoptium/aqa-tests:master'], nargs='+')
+    parser.add_argument('--openj9_repo', default=['eclipse-openj9/openj9:master'], nargs='+')
+    parser.add_argument('--tkg_repo', default=['adoptium/TKG:master'], nargs='+')
+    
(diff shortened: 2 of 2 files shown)


llxia (approved):
LGTM

sophia-guo:
Simply removing the logic of parsing file runAqaConfig.json  and adding argument directly in AqaArgParse.py may not work properly as runAqaArgParse.py is used in different repos ( aqa-tests, openj9, etc.)

sophia-guo:
The current change should work for TKG repo. 

To support other repos two steps are needed:

1. tell AqaArgParse.py, which repo the workflow is using
2. Add arguments accordingly, sudo code:

```
if ( TKGrepo ){
    parser.add_argument('--openjdk_testrepo', default=['adoptium/aqa-tests:master'], nargs='+')
    parser.add_argument('--openj9_repo', default=['eclipse-openj9/openj9:master'], nargs='+')
} else if (aqa-tests) {
    parser.add_argument('--openj9_repo', default=['eclipse-openj9/openj9:master'], nargs='+')
    parser.add_argument('--tkg_repo', default=['adoptium/TKG:master'], nargs='+')
} if (openj9) {
    parser.add_argument('--openjdk_testrepo', default=['adoptium/aqa-tests:master'], nargs='+')
    parser.add_argument('--tkg_repo', default=['adoptium/TKG:master'], nargs='+')
}

```

llxia:
@sophia-guo can you share the reason why we need to check the repo for adding arguments? Also, is there a good way of testing this?

llxia:
ping @sophia-guo

sophia-guo:
The current story for PR comment triggers is we won't pass in PR's repo and branch to runAqaArgParse.py. The functionality of the file is to parse explicit parameters set by user's comment. PR's repo and branch can be easily set by github action API, which don't need to set by comment and be parsed to set as environment variable. All those parameters will be consumed by run-aqa action. 

Alternatively we can just add all parameters without checking repos and correspondingly update all workflow files in different repos and readme. @llxia thoughts?

sophia-guo:
Original PR to add this json file https://github.com/adoptium/aqa-tests/pull/2465

sophia-guo:
PR repo information is explicitly setting by github api 
[         tkg_Repo: '${{ fromJSON(steps.get-pr.outputs.result).head.repo.full_name }}:${{ fromJSON(steps.get-pr.outputs.result).head.ref }}'](https://github.com/adoptium/TKG/blob/master/.github/workflows/runAqa.yml#L158), which is not needed to be parsed by AqaArgParse.py. It might not be the best solution but it doesn't hurt to set it as redundant parameter at the moment.

sophia-guo (approved):

PR #263, opened 2022-01-24
Account for java.security.manager behaviour changes in JDK18

babsingh:
In JDK18+, `java.security.manager == null` behaves as
`-Djava.security.manager=disallow`.

In JDK17-, `java.security.manager == null` behaves as
`-Djava.security.manager=allow`.

For OpenJ9 tests to work as expected, `-Djava.security.manager=allow` behaviour
is needed in JDK18+.

Signed-off-by: Babneet Singh <sbabneet@ca.ibm.com>

### DIFF
--- a/openj9Settings.mk
+++ b/openj9Settings.mk
@@ -126,10 +126,21 @@ else
 	ADD_JVM_LIB_DIR_TO_LIBPATH:=export $(TEST_LIB_PATH);
 endif
 
+#######################################
+# In JDK18+, java.security.manager == null behaves as -Djava.security.manager=disallow.
+# In JDK17-, java.security.manager == null behaves as -Djava.security.manager=allow.
+# For OpenJ9 tests to work as expected, -Djava.security.manager=allow behaviour is
+# needed in JDK18+.
+#######################################
+ifeq (18,$(word 1,$(sort 18 $(JDK_VERSION))))
+	EXTRA_OPTIONS += -Djava.security.manager=allow
+endif
+
 ifneq ($(DEBUG),)
 $(info JAVA_SHARED_LIBRARIES_DIR is set to $(JAVA_SHARED_LIBRARIES_DIR))
 $(info VM_SUBDIR is set to $(VM_SUBDIR))
 $(info ADD_JVM_LIB_DIR_TO_LIBPATH is set to $(ADD_JVM_LIB_DIR_TO_LIBPATH))
+$(info EXTRA_OPTIONS is set to $(EXTRA_OPTIONS))
 endif
 
 #######################################


babsingh:
### Javadoc References
- **JDK17:** https://cr.openjdk.java.net/~iris/se/17/spec/latest/api/java.base/java/lang/SecurityManager.html#java.security.manager
- **JDK18:** https://cr.openjdk.java.net/~iris/se/18/spec/latest/api/java.base/java/lang/SecurityManager.html#java.security.manager

This PR is required before https://github.com/eclipse-openj9/openj9/pull/14329 can be merged.

fyi @renfeiw @llxia @JasonFengJ9

openj9Settings.mk
```diff
@@ -126,10 +126,21 @@ else
...
 endif
 
+#######################################
+# In JDK18+, java.security.manager == null behaves as -Djava.security.manager=disallow.
+# In JDK17-, java.security.manager == null behaves as -Djava.security.manager=allow.
+# For OpenJ9 tests to work as expected, -Djava.security.manager=allow behaviour is
+# needed in JDK18+.
+#######################################
+ifeq (18,$(word 1,$(sort 18 $(JDK_VERSION))))
+	EXTRA_OPTIONS += -Djava.security.manager=allow
```
renfeiw, line 136: Just gave a second thought, this might cause issue because the options don't have space in between. Maybe this is better:
```
EXTRA_OPTIONS := $(EXTRA_OPTIONS) -Djava.security.manager=allow
```

babsingh: As per the documentation (https://www.gnu.org/software/make/manual/html_node/Appending.html), both approaches are identical.

renfeiw (approved):

babsingh:
> Is this OpenJ9 specific or something that should be more generally applied?

The `java.security.manager` behaviour change will be seen in every JDK18+ implementation. So, it is not OpenJ9 specific.

Also, I noticed that `EXTRA_OPTIONS` are appended to the end of the JVM arguments. This will override any identical test or user specific JVM arguments. Is it possible to prepend `-Djava.security.manager=allow` to the JVM arguments?

openj9Settings.mk
```diff
@@ -126,10 +126,21 @@ else
...
+endif
```
llxia, line 137: The above code does not seem to work as expected:

1. it sets `-Djava.security.manager=allow` for JDK8 
2. `-Djava.security.manager=allow` is added multiple times. See JVM_OPTIONS.

JDK8: https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/564/console
```
11:28:52  ===============================================
11:28:52  Running test floatSanityTests_2 ...
11:28:52  ===============================================
11:28:52  floatSanityTests_2 Start Time: Tue Jan 25 11:24:08 2022 Epoch Time (ms): 1643127848273
11:28:52  "/home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-image/bin/java" -Xshareclasses:destroyAll; "/home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-image/bin/java" -Xshareclasses:groupAccess,destroyAll; echo "cache cleanup done";
11:28:52  JVMSHRC005I No shared class caches available
11:28:52  JVMSHRC005I No shared class caches available
11:28:52  cache cleanup done
11:28:52  variation: Mode150
11:28:52  JVM_OPTIONS:  -XX:+UseCompressedOops -Xint -Djava.security.manager=allow -Djava.security.manager=allow -Djava.security.manager=allow -Djava.security.manager=allow
11:28:52  { itercnt=1; \
11:28:52  mkdir -p "/home/jenkins/workspace/Grinder/aqa-tests/TKG/../TKG/output_16431278457721/floatSanityTests_2"; \
11:28:52  cd "/home/jenkins/workspace/Grinder/aqa-tests/TKG/../TKG/output_16431278457721/floatSanityTests_2"; \
11:28:52  "/home/jenkins/workspace/Grinder/openjdkbinary/j2sdk-image/bin/java"  -XX:+UseCompressedOops -Xint -Djava.security.manager=allow -Djava.security.manager=allow -Djava.security.manager=allow -Djava.security.manager=allow \
11:28:52  -cp "/home/jenkins/workspace/Grinder/aqa-tests/TKG/../../jvmtest/TestConfig/resources:/home/jenkins/workspace/Grinder/aqa-tests/TKG/../TKG/lib/testng.jar:/home/jenkins/workspace/Grinder/aqa-tests/TKG/../TKG/lib/jcommander.jar:/home/jenkins/workspace/Grinder/aqa-tests/TKG/../../jvmtest/functional/Java8andUp/GeneralTest.jar" \
11:28:52  org.testng.TestNG -d "/home/jenkins/workspace/Grinder/aqa-tests/TKG/../TKG/output_16431278457721/floatSanityTests_2" "/home/jenkins/workspace/Grinder/aqa-tests/TKG/../../jvmtest/functional/Java8andUp/testng.xml" \
11:28:52  -testnames floatSanityTests \
11:28:52  -groups level.sanity \
11:28:52  -excludegroups d.*.linux_x86-64,d.*.arch.x86,d.*.os.linux,d.*.bits.64,d.*.generic-all; \
11:28:52  if [ $? -eq 0 ] ; then echo ""; echo "floatSanityTests_2""_PASSED"; echo ""; cd /home/jenkins/workspace/Grinder/aqa-tests/TKG/..; rm -f -r "/home/jenkins/workspace/Grinder/aqa-tests/TKG/../TKG/output_16431278457721/floatSanityTests_2"; else echo ""; echo "floatSanityTests_2""_FAILED"; echo ""; fi; } 2>&1 | tee -a "/home/jenkins/workspace/Grinder/aqa-tests/TKG/../TKG/output_16431278457721/TestTargetResult";
11:28:53  Exception in thread "main" java/lang/InternalError: Could not create SecurityManager: allow
11:28:53  	at sun/misc/Launcher.<init> (Launcher.java:106)
11:28:53  	at sun/misc/Launcher.<clinit> (Launcher.java:54)
11:28:53  	at java/lang/ClassLoader.initializeClassLoaders (ClassLoader.java:196)
11:28:53  	at java/lang/Thread.initialize (Thread.java:405)
11:28:53  	at java/lang/Thread.<init> (Thread.java:154)
11:28:53  
11:28:53  floatSanityTests_2_FAILED
```

JDK18: https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/568/console
```
11:43:54  JVM_OPTIONS:  -Xint -Djava.security.manager=allow -Djava.security.manager=allow -Djava.security.manager=allow -Djava.security.manager=allow
```

pshipton:
If you are running functional or system tests that originate from OpenJ9/IBM on jdk18 that still depend on the security manager, then this option is needed. Afaik we haven't started splitting these tests out so they can be excluded on versions were they can never work (jdk19+?), or explicitly run with this option on jdk18.

We don't want this option when running OpenJDK testing.

pshipton:
If `-Djava.security.manager=allow` is prepended, likely this is sufficient to keep everything working. Prepending should be fine even for OpenJDK testing (for jdk18). Prepending can be done by setting an environment variable such as `OPENJ9_JAVA_OPTIONS `, but this only works on OpenJ9. We'd have to check if all the testing can still work with this environment variable set.

smlambert:
Ouch, this sounds somewhat complicated.

If you do not want this option for openjdk testing, and only need it for system and functional tests (and not the other categories of test), then TKG may not the correct location to make this change, but rather in [systemtest.mk](https://github.com/adoptium/aqa-tests/blob/master/system/systemtest.mk) and functional.mk (which we should consider creating to handle such cases).

pshipton:
Well any testing that expects old security manager behaviour is affected. Testing related to OpenJDK would have been fixed to expect the change in behaviour. This is an attempt at a workaround to put off having to fix all the OpenJ9 testing. Any tests that set a security manager need to be separated so (a) they can enable the security manager on jdk18 (because it's disabled by default) (b) the tests can be disabled on a future jdk version where the security manager is removed.

There is probably some small amount of testing that sets `-Dsecurity.manager=<something>` on the command line, and this needs to take affect rather than the option in this PR.

smlambert:
Thanks for the clearer explanation (in https://github.com/adoptium/TKG/pull/263#issuecomment-1021462298), this is a temporary change to be done in the simplest way possible while waiting for an open issue (somewhere), which describes a review and update of test material.  However this change is accomplished, the comment should probably reflect that its temp code and what issue needs to be addressed in order to remove it.

babsingh:
**Summarizing requirements + questions:**
- This change will go in `aqa-tests` ... `systemtest.mk` and `functional.mk` since we only want to influence OpenJ9 testing.
- The system property needs to prepended using `OPENJ9_JAVA_OPTIONS`. Do we need an OpenJDK equivalent?
- There needs to be an issue to segregate + disable security manager tests for JDK18+. Is there already an existing issue?
- There needs to be a note to undo the temporary change pointing to the above issue.

@smlambert @pshipton can you please confirm if all the requirements are accounted for?

pshipton:
Afaik the "OpenJ9" functional and system tests are used to test OpenJDK by Adoptium, so a perfect solution needs to keep these working.

I'm not aware of any specific issue to segregate the security manager tests. The reason for doing this is because the security manager will be removed completely in some future release so adding a command option option will not work to fix them. The tests should always be run on jdk18 and releases that don't have the security manager removed with `-Djava.security.manager=allow` (except when the test already specifies `-Djava.security.manager`). This can be done (hopefully) via global changes, or directly in the segregated test cases.

pshipton:
And of course we want to continue running the tests as-is on jdk8, 11, 17.

llxia:
> When I run the OpenJ9 tests locally, I only clone the TKG directory in openj9/test/TKG; aqa-tests is not involved. How will the changes in aqa-tests -> systemtest.mk and functional.mk transfer to the TKG directory?

For the local run, please `git clone https://github.com/adoptium/aqa-tests.git` to get all tests (Plz refer https://github.com/eclipse-openj9/openj9/wiki/Reproducing-Test-Failures-Locally for details)

babsingh:
I scanned the below repos:
- https://github.com/adoptium/aqa-tests
- https://github.com/adoptium/TKG
- https://github.com/adoptium/aqa-systemtest
- https://github.com/eclipse-openj9/openj9

Only the OpenJ9 repo has tests which use `setSecurityManager`.

Closing this PR. I will be making isolated changes targeting impacted tests in the OpenJ9 repo. https://github.com/eclipse-openj9/openj9/pull/14329: second commit covers test changes.

Issue #260, opened 2022-01-22
Create a metadata file for AQAvit certification use case


smlambert:
As part of the AQAvit certification use case, vendors wishing to certify need to provide both the test results in the form of TAP files (which are already produced by TKG), and metadata that contains information about the "object under test" (the JDK binary).

We already print some of this information to console, but do not capture it into a file for easy archiving.

This issue is to both discuss and determine the initial required set of information to capture, but also to then enhance TKG to produce the metadata file during a test run.

Information that should be included are:
- java -version output
- information regarding the tags/SHAs of the 'object under test' - for OpenJ9 this information is in the java -version output, for hotspot, we have to extract it by other means

llxia:
The issue for Archive the metadata file is created at aqa-tests repo: https://github.com/adoptium/aqa-tests/issues/3264

JeromeJu:
Would like to start on this :)

llxia:
Release info: https://github.com/adoptium/aqa-tests/blob/master/get.sh#L596-L601

PR #266, opened 2022-02-02
Add java -version output

JeromeJu:
Add 'java -version' to the metadata file.

Fixes: adoptium/TKG#260

Signed-off-by: Jerome Ju jeromeqju@gmail.com

### DIFF
--- a/src/org/openj9/envInfo/EnvDetector.java
+++ b/src/org/openj9/envInfo/EnvDetector.java
@@ -74,6 +74,8 @@ public class EnvDetector {
 			output.write(JDKVERSIONvalue);
 			output.write(JDKIMPLvalue);
 			output.write(JDKVENDORvalue);
+			output.close();
+			output = new BufferedWriter(new OutputStreamWriter(new FileOutputStream("AQACert.log")));
 			output.write(JavaVersionValue);
 			output.close();
 		} catch (IOException e) {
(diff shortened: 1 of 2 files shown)


JeromeJu:
What info included in the result of executing 'java -version'shall we kept out of the writing into file?

JeromeJu:
How can I test this before merging?

llxia:
You can test in Grinder using `TKG_OWNER_BRANCH`: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/build?delay=0sec

image

src/org/openj9/envInfo/EnvDetector.java
```diff
@@ -72,6 +74,7 @@ private static void getJavaInfo() {
 			output.write(JDKVERSIONvalue);
 			output.write(JDKIMPLvalue);
 			output.write(JDKVENDORvalue);
+			output.write(JavaVersionValue);
```
llxia, line 77: We need to output into a different file - AQACert.log

llxia:
@JeromeJu  Could you double-check where the file is output to? It is expected this log file is output under TKG dir https://github.com/adoptium/aqa-tests/pull/3316. Could you also provide the Grinder link? Thanks

JeromeJu:
I tried running the Grinder test earlier but not fully sure how to check the file output on Jenkins.
Grinder link [#3539](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3539/console)

Noticed that the `No artifacts found that match the file pattern "aqa-tests/TKG/AQACert.log". Configuration error?` warning is gone. So I guessed it shall output to the correct path? as discussed in https://github.com/adoptium/aqa-tests/pull/3316.

llxia:
> So I guessed it shall output to the correct path?

yes, just add `pwd` and see where is your current dir. 
@renfeiw any variable we can directly use to get to TKG dir?

llxia:
Thanks @JeromeJu . The file is generated at the correct place.

llxia (approved):
LGTM

smlambert (approved):
thanks @JeromeJu !

PR #267, opened 2022-02-02
Add feature into TKG

### LINKED ISSUE #268: Create a feature tag for test
We need a mechanism to control a test against a JDK feature, for example AOT, CRIU, etc. We already had some tags such as `<aot>` to control tests, but it only works for aot. We need a generalized way to handle any upcoming feature. This will replace the `<aot>` tag.

Here is the result of the work explained with AOT as an example, we can define one of the following options for a test. For other features, just replace AOT with the feature name that's defined in TEST_FLAG. Multiple `<feature>` can be enclosed in one `<features>`.
```
<features>
           <feature>AOT:applicable</feature> ------ This is the default option. The test will run whether or not TEST_FLAG contains 
AOT (Special treatment will be applied when TEST_FLAG contains AOT, it will be run multiple times and special JVM option will be 
applied)
           <feature>AOT:nonapplicable</feature> ------ The test will NOT run when TEST_FLAG contains AOT
           <feature>AOT:explicit</feature> ------ The test will run whether or not TEST_FLAG contains AOT (No special treatment will 
be applied when TEST_FLAG contains AOT, the test will run as-is) Currently, this option is only meaningful for AOT.
           <feature>AOT:required</feature> ------ the test will run ONLY when TEST_FLAG contains AOT (Special treatment will be 
applied)
</features>
```

renfeiw:
- supports turn on/off test based on feature
- combine existing <aot> tag with feature tag

Fixes #268 

### DIFF
--- a/resources/playlist.xsd
+++ b/resources/playlist.xsd
@@ -54,6 +54,13 @@
               <xs:element type="xs:string" name="platformRequirements" maxOccurs="1" minOccurs="0"/>
               <xs:element type="xs:string" name="platform" maxOccurs="1" minOccurs="0"/>
               <xs:element type="xs:string" name="aot" maxOccurs="1" minOccurs="0"/>
+              <xs:element name="features" maxOccurs="1" minOccurs="0">
+                <xs:complexType>
+                  <xs:sequence>
+                    <xs:element type="xs:string" name="feature" maxOccurs="unbounded" minOccurs="1"/>
+                  </xs:sequence>
+                </xs:complexType>
+              </xs:element>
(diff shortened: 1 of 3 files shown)


renfeiw:
Internal grinder: Grinder_TKG/1498, compared with grinder without change Grinder_TKG/1497. Checked test with different aot settings.

renfeiw:
This PR needs to merge together with https://github.com/eclipse-openj9/openj9/pull/14417

llxia (approved):
Thanks @renfeiw

smlambert (approved):
👍

Issue #262, opened 2022-01-24
TKG does not set correct spec for x86-64_alpine-linux


llxia:
TKG does not set correct spec for x86-64_alpine-linux (below `--spec=linux_x86-64` is used)
https://ci.adoptopenjdk.net/job/Test_openjdk18_hs_sanity.openjdk_x86-64_alpine-linux/53/consoleText
```
[2022-01-15T00:38:47.131Z] /home/jenkins/workspace/Test_openjdk18_hs_sanity.openjdk_x86-64_alpine-linux/openjdkbinary/j2sdk-image/bin/java -cp "./bin/TestKitGen.jar:./lib/json-simple.jar" org.testKitGen.MainRunner --mode=tests --spec=linux_x86-64 --jdkVersion=18 --impl=hotspot --vendor="eclipse" --buildList=openjdk --iterations=1 --testFlag= --testTarget=sanity.openjdk --testList= --numOfMachines= --testTime= --TRSSURL= 
```

TKG AUTO_DETECT needs to be updated to support this. https://github.com/adoptium/TKG/blob/master/src/org/openj9/envInfo/JavaInfo.java#L27

Below is the code for checking alpine-linux:
https://github.com/adoptium/aqa-tests/blob/master/jck/jtrunner/JavaTestRunner.java#L1235

JeromeJu:
I'd like to confirm with the testing, shall I set the `PLATFORM` parameter in Grinder to `x86-64_alpine-linux`?

llxia:
Yes, that is correct.

PR #271, opened 2022-02-08
Add spec for alpine_linux

JeromeJu:
Add spec for alpine_linux

Fixes https://github.com/adoptium/TKG/issues/262

Signed-off-by: Jerome Ju([jeromeqju@gmail.com](mailto:jeromeqju@gmail.com))

### DIFF
--- a/src/org/openj9/envInfo/JavaInfo.java
+++ b/src/org/openj9/envInfo/JavaInfo.java
@@ -14,6 +14,10 @@
 
 package org.openj9.envInfo;
 
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.Files;
+
 public class JavaInfo {
 
     public String getSPEC(String javaImplInfo) {
@@ -25,6 +29,10 @@ public class JavaInfo {
         System.out.println("System.getProperty('java.fullversion')=" + fullversion + "\n");
         String spec = "";
         if (osName.contains("linux")) {
+            Path alpine = Paths.get("/etc/alpine-release");
+			if (Files.exists(alpine)) {
+				spec = "linux_x86-64";
+			}
             spec = "linux";
         } else if (osName.contains("win")) {
             spec = "win";
--- a/src/org/openj9/envInfo/JavaInfo.java
+++ b/src/org/openj9/envInfo/JavaInfo.java
@@ -31,7 +31,7 @@ public class JavaInfo {
         if (osName.contains("linux")) {
             Path alpine = Paths.get("/etc/alpine-release");
 			if (Files.exists(alpine)) {
-				spec = "linux_x86-64";
+				spec = "alpine-linux_x86-64";
 			}
             spec = "linux";
         } else if (osName.contains("win")) {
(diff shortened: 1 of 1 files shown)


JeromeJu:
The format is a bit off while viewing on Git

src/org/openj9/envInfo/JavaInfo.java
```diff
@@ -25,6 +29,10 @@ public String getSPEC(String javaImplInfo) {
         System.out.println("System.getProperty('java.fullversion')=" + fullversion + "\n");
         String spec = "";
         if (osName.contains("linux")) {
+            Path alpine = Paths.get("/etc/alpine-release");
+			if (Files.exists(alpine)) {
+				spec = "linux_x86-64";
+			}
```
llxia, line 35: - The logic is not correct. If exists(alpine), the spec is `linux_x86-64`? 
Please see example https://github.com/adoptium/aqa-tests/blob/master/jck/jtrunner/JavaTestRunner.java#L1240
- spec will be appended with arch and bit information in the logic below. We only set os here (i.e., `alpine-linux`), not hardcode the whole spec.
- please run a Grinder after the code change.

JeromeJu:
In Grinder run [#3535](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/3535/console) 
Encountered `23:49:09  get jdk binary...
23:49:09  _ENCODE_FILE_NEW=UNTAGGED curl -OLJSks  https://api.adoptium.net/v3/binary/latest/8/ea/alpine-linux/x64/jdk/hotspot/normal/adoptium?project=jdk
23:49:09  Uncompressing file: adoptium?project=jdk ...
23:49:09  gzip: invalid magic
23:49:09  tar: short read`

Would like to confirm if it's the issue related to [Alpine version issue](https://unix.stackexchange.com/questions/302192/how-to-solve-tar-invalid-magic-error-on-linux-alpine) or is it that I've set up the wrong params for the build?

JeromeJu:
Added the change in [#3328](https://github.com/adoptium/aqa-tests/pull/3328)

> > https://github.com/adoptium/aqa-tests/blob/master/buildenv/jenkins/openjdk_tests#L74
> 
> Yes I agree, `alpine-linux_x86-64` would be better.

src/org/openj9/envInfo/JavaInfo.java
JeromeJu: Thanks for the correction. I'd like to ask should the `osName.equals("linux")` be equivalent to `osName.contains("linux")` here? 

Also, may I ask what's the reason for changing the spec to `alpine-linux_x86-64`?

renfeiw: >I'd like to ask should the osName.equals("linux") be equivalent to osName.contains("linux") here?

Not sure if osName contains extra strings other than linux, I guess contains is better than equals.

> Also, may I ask what's the reason for changing the spec to alpine-linux_x86-64?

You will also need to update ottawa.csv file, this file controls which spec runs on what platform for what JVM option. Add a column of  alpine-linux_x86-64 in it. We use different spec to run or skip the test based on ottawa.csv.

Two comments on your current change:
- use space instead of tab for indentation so it follows the existing format
- you need to have the `spec="linux"` in else condition

JeromeJu:
Closed with a new PR #273  in draft due to branch name formatting.

PR #291, opened 2022-03-03
Add Mode189 to test -Xtune:throughput

cjjdespres:
Fixes: eclipse-openj9/openj9/issues/14551
Signed-off-by: Christian Despres <despresc@ibm.com>

### DIFF
--- a/resources/modes.xml
+++ b/resources/modes.xml
@@ -1810,6 +1810,15 @@
 			</setting>
 		</settings>
 	</mode>
+	<mode number="189">
+		<description>JIT tune throughput</description>
+		<settings>
+			<setting type="either">
+				<clArg>-Xtune:throughput</clArg>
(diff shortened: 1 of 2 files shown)


cjjdespres:
Attn @mpirvu

llxia:
As suggested in https://github.com/eclipse-openj9/openj9/issues/14551#issuecomment-1049251410, we would prefer to use `-Xtune:throughput` directly, not adding a new Mode. It is not necessary to create a Mode for a single jvm option that can run on all platforms. It will cause more maintenance. May I know the reason why this is needed?

cjjdespres:
I don't think it's absolutely needed. I added it because of https://github.com/eclipse-openj9/openj9/issues/14551#issuecomment-1049234938, but I think we're both okay with it not being a new mode.

cjjdespres:
If we go that way then this PR can be closed and I'll change the other PR so it uses the option directly.

Issue #295, opened 2022-03-11
Update the --openjdk_testrepo flag name in runAqaArgParse.py


JeromeJu:
Describe the bug
In the runAqaArgParse.py file, the --openjdk_testrepo arg name should be updated to --aqa-testrepo

Expected behavior
The run-aqa should later take in --aqa-testrepo flag

JeromeJu:
Hi, I was investigating the effect of the one-liner in #298 via the  [similar workflow](https://github.com/JeromeJu/TKG/runs/5553598711?check_suite_focus=true)

JeromeJu:
I checked again to make sure the checked out repo is the intended one. However, it still would not bring the correct output. As suggested from last meeting, shall we try merging this one-liner?

PR #297, opened 2022-03-11
Add condition for runAqa config Json file

JeromeJu:
Added condition checks for runAqa Json file. Skip if there's no such file.

Fixes: #219 

Signed-off-by: Jerome Ju [jeromeqju@gmail.com](mailto:jeromeqju@gmail.com)

PR #298, opened 2022-03-11
Change arg name to aqatest

JeromeJu:
Change arg name to aqatest for runaqaArgParse script.

Will finish off with testing with the same workflow if PR #297 resolves.

Fixes: #295 

Signed-off-by: Jerome Ju [jeromeqju@gmail.com](mailto:jeromeqju@gmail.com)


JeromeJu:
https://github.com/JeromeJu/TKG/runs/5553660477?check_suite_focus=true

JeromeJu:
I checked again to make sure the checked out repo is the intended one. However, it still would not bring the correct output. As suggested from last meeting, shall we try merging this one-liner? 

I can revert and investigate if needed.

PR #320, opened 2022-05-17
Add MALLOCOPTIONS= in environment variable

renfeiw:

### DIFF
--- a/envSettings.mk
+++ b/envSettings.mk
@@ -71,3 +71,4 @@ else
     endif
 endif
 
+export MALLOCOPTIONS=


pshipton:
The reason for this change is that OpenJ9 builds for AIX sets `MALLOCOPTIONS=multiheap,considersize` when the JVM is started, and this gets inherited by ant, which calls git, and then causes the following failure:
```
00:20:42  getOpenjdk:
00:20:42       [echo]  Using user specified repo and branch
00:20:42       [echo] git clone --depth 1 -q -b openj9 git@github.com:ibmruntimes/openj9-openjdk-jdk8.git openjdktemp
00:21:17       [exec] fatal: pack has bad object at offset 71306024: inflate returned -5
00:21:17       [exec] fatal: fetch-pack: invalid index-pack output
00:21:17       [exec] Result: 128
00:21:17       [echo] git clone -q git@github.com:ibmruntimes/openj9-openjdk-jdk8.git openjdktemp 
00:21:31       [exec] fatal: pack has bad object at offset 232035628: inflate returned -5
00:21:31       [exec] fatal: fetch-pack: invalid index-pack output
```

renfeiw:
tested in Grinder/23820

smlambert (approved):

llxia:
Related: infrastructure/issues/6608

PR #322, opened 2022-05-20
add support for Alpine Linux aarch64

gdams:


karianna (approved):
LGTM

llxia (approved):
LGTM

smlambert:
Does there need to be an entry in modes file?  Perhaps hold off on merging so we can run a Grinder to see how it goes.

smlambert:
That Grinder failed due to other issues.  I am will merge this as it is not incorrect, and if we need additional updates in TKG it can come as another PR.

PR #339, opened 2022-08-03
Create default spec

### LINKED ISSUE #341: Add a default spec
Most platforms do not need their own spec and would use the same set of [modes](https://github.com/adoptium/TKG/blob/master/resources/modes.xml) (standard set of command line options).  

Having a default spec, simplifies [ottawa.csv](https://github.com/adoptium/TKG/blob/master/resources/ottawa.csv) and will make it easier to add new platforms and also mean less need to change that file (which is awkward to change and a carry over from some old tooling that we have not removed ourselves from just yet).

renfeiw:
Fixes #341 
- add default modes in ottawa.csv
- when platform not found, fall back to default


renfeiw:
Modified auto detection logic to test default case. 64-bits: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5356 32-bits: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5357

renfeiw:
test on non-default case: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5385

llxia (approved):
LGTM

smlambert (approved):
LGTM, thanks for the Grinder links and proactive testing!

Issue #342, opened 2022-08-09
make compile fails for openjdk && system on Windows


j-braley:
After installing `aqa-tests` and running `./compile.sh` or `make compile` (in TKG/) make exits with `Error 2`.
The specific stage where I am observing these failures is the call to `/TKG/scripts/getFileWithRetry.sh` and the error is `No such file or directory`

Steps to reproduce:
1. Launch up Cygwin
2. Export env vars
3. Clone `aqa-tests`
4. Run `./get.sh`
5. Run `./compile.sh`

Expected that make completes with no error.
Actual make exits with error code 2.
Output below from running BUILD_LIST=openjdk but I have seen failures for system as well.
```

$ export TEST_JDK_HOME=/path/to/my/test/jdk
$ export TESTIMAGE_PATH=/path/to/test/image
$ export JDK_VERSION=11
$ export USE_TESTENV_PROPERTIES=true
$ export BUILD_LIST=openjdk

$ ./get.sh
TESTDIR: /home/joebraley/test-breakage/aqa-tests
load ./testenv/testenv.properties
get testKitGen...
git clone -q https://github.com/adoptium/TKG.git
git rev-parse v0.9.3
git checkout -q -f f614fc5ebe681000911571945fe59f85665ef40c
check adoptium repo and TKG repo SHA
/home/joebraley/test-breakage/aqa-tests/TKG/scripts/getSHA.sh --repo_dir /home/joebraley/test-breakage/aqa-tests --output_file /home/joebraley/test-breakage/aqa-tests/TKG/SHA.txt
Check sha in /home/joebraley/test-breakage/aqa-tests and store the info in /home/joebraley/test-breakage/aqa-tests/TKG/SHA.txt
touch /home/joebraley/test-breakage/aqa-tests/TKG/SHA.txt
================================================
timestamp: 20220809-164630
repo dir: /home/joebraley/test-breakage/aqa-tests
git repo:
  Fetch URL: https://github.com/adoptium/aqa-tests.git
sha:
fdc28ada41889a73adf8bb5eabcd15d9c058df06
/home/joebraley/test-breakage/aqa-tests/TKG/scripts/getTestenvProperties.sh --repo_dir /home/joebraley/test-breakage/aqa-tests --output_file /home/joebraley/test-breakage/aqa-tests/testenv/testenv.properties --repo_name ADOPTOPENJDK
USE_TESTENV_PROPERTIES was set, not writing to testenv.properties
[...]

PR #343, opened 2022-08-10
feat: make sure for both cygwin and msys cygpath is used

zdtsw:


llxia:
Please provide the Grinder link. Thanks

smlambert:
What issue does this address?

zdtsw:
for x86_64-linux https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5417
for x86-32_windows https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5418
for x86-64_mac https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5419
for ppc64_aix https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5420

> What issue does this address?

when I was checking why run-aqa/aqa-tests not working on windows-2019, i found the code in temurin-build https://github.com/adoptium/temurin-build/blob/0aee022f86093836998a48b9795002fa6afeb817/sbin/prepareWorkspace.sh#L420 checks on both cygwin and mingw convert path. I do not know if this could be problem if we use a newer version or anyone wants to use AQA but install with mingw on their windows, just think it is good to align the logic in both code.

settings.mk
```diff
@@ -111,17 +111,21 @@ endif
...
 		CYGWIN:=0
 	endif
 endif
 
+$(info CYGWIN is $(CYGWIN))
 ifeq ($(CYGWIN),1)
 	TEST_ROOT := $(shell cygpath -w $(TEST_ROOT))
 endif
 TEST_ROOT := $(subst \,/,$(TEST_ROOT))
+$(info TEST_ROOT is $(TEST_ROOT))
```
renfeiw, line 128: I don't think we need to print out TEST_ROOT, it is already printed here: 	https://github.com/adoptium/TKG/blob/15ae4d0746c3fc95564266e156bc57deae8d985a/settings.mk#L276. 
As for `CYGWIN`, if you want need to keep it, please move it to the same place. Using `info` here would cause it to be printed multiple times. Also it will be more meaningful to print something like `$(OSTYPE)` instead of just `CYGWIN is 1`.

zdtsw: thanks @renfeiw 
removed both "info" lines

llxia:
@zdtsw what error did you encounter on windows-2019? Did this PR resolve the problem?

zdtsw:
> @zdtsw what error did you encounter on windows-2019? Did this PR resolve the problem?

the error can be found from https://github.com/adoptium/run-aqa/runs/7789840940?check_suite_focus=true it is on 2022/latest.
two issues are reported for this https://github.com/adoptium/TKG/issues/342 and https://github.com/adoptium/run-aqa/issues/133

the PR is not for fix the problem, but only to match what we have in the build script's logic to support both cygwin and minGW.

smlambert:
Given cygwin is a stated [prereq](https://github.com/adoptium/aqa-tests/blob/master/doc/Prerequisites.md#prerequisites) for AQAvit, this change is likely not warranted.

zdtsw:
> Given cygwin is a stated [prereq](https://github.com/adoptium/aqa-tests/blob/master/doc/Prerequisites.md#prerequisites) for AQAvit, this change is likely not warranted.

right! this PR probably not needed for the current requirement. 
what I am worrying is, cygwin might not work with later windows version, if that is the case(probably as one item to investigate) then it either be limited to 2016 or need re-write logic of cygwin installation to something else or use wsl.  https://cygwin.com/pipermail/cygwin-announce/2022-January/010438.html

smlambert:
Yes, in which case we should create an issue and discuss and design an approach, not just needle in a PR here or there without thinking of all of the implications (and testing that will need to happen) if we want to support a new environment.

Issue #336, opened 2022-07-11
Support machine info detection in TKG


llxia:
Some tests may only run on certain micro-architecture. TKG should check micro-architecture on the machine and execute/skip tests accordingly. Frome example, some VectorAPI tests only run on Z13/Z14 or newer, CRIU portability tests run on skylake, etc. This feature may further extend to check other machine info (i.e., RAM, CPU, application installed, etc) as needed.

TKG runs cmd to detect the machine info:

- cmd for detecting skylake
```
cat /proc/cpuinfo | grep 'model name' | uniq
```
- cmd for detecting z version with newer openj9 JDK:
```
$ ./semaru/bin/java -Xshareclasses:none -Xjit:verbose -version 2>&1 | grep Name:
#INFO:          Name: z15
```

`platformRequirements ` in playlist.xml should be enhanced for setting specific hardware or software requirements.

Related issue:
runtimes/openj9-jit-z/issues/711
runtimes/infrastructure/issues/7037
https://github.com/adoptium/TKG/issues/45#issuecomment-1167750455

smlambert:
Would also like a check for whether we are in a containerized environment or not, something like this:
`if grep -sq 'docker\|lxc' /proc/1/cgroup; then    echo "I am running on Docker."; else echo "not docker "; fi`

smlambert:
It seems like both the support for microArch and a PR for docker info was merged.  Can this issue be closed?  

Noting a separate issue was raised to detect Podman versus Docker https://github.com/adoptium/TKG/issues/503, so I would not hold this issue open for that.

llxia:
Issue https://github.com/adoptium/TKG/issues/471 is for z versions.

PR #351, opened 2022-09-06
Adding docker info into env detection

renfeiw:

### DIFF
--- a/src/org/openj9/envInfo/MachineInfo.java
+++ b/src/org/openj9/envInfo/MachineInfo.java
@@ -38,6 +38,7 @@ public class MachineInfo {
 	public static final String[] CPU_CORES_CMD_SOLARIS = new String[] {"bash", "-c", "psrinfo | wc -l"};
 	public static final String[] NUMA_CMD = new String[] {"bash", "-c", "numactl --show | grep 'No NUMA support available on this system"};
 	public static final String[] SYS_VIRT_CMD = new String[] {""};
+	public static final String[] CHECK_DOCKER_CMD = new String[] {"bash", "-c", "if grep -sq 'docker\\|lxc' /proc/1/cgroup; then echo 'running on docker'; else echo 'not on docker'; fi"};
 
 	// Software
 	public static final String[] SYS_OS_CMD = new String[] {"uname", "-s"};
@@ -189,6 +190,7 @@ public class MachineInfo {
 		putInfo(new Info("microArch", MICRO_ARCH_CMD, microArch, null));
 		putInfo(new Info("sysOS", SYS_OS_CMD, ce.execute(SYS_OS_CMD), null));
 		putInfo(new Info("ulimit", ULIMIT_CMD, ce.execute(ULIMIT_CMD), null));
+		putInfo(new Info("docker", CHECK_DOCKER_CMD, ce.execute(CHECK_DOCKER_CMD), null));
 	}
 	
 	private void getPrerequisiteInfo() {


renfeiw:
related: https://github.com/adoptium/TKG/issues/336#issuecomment-1196659321

renfeiw:
grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5634/

smlambert (approved):
Ran a few additional grinders to cover primaries [win](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5636/)/[mac](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5637/) and the platform where utilities is the most restricted in terms of options / [solaris](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5635/)).  

Solaris hits this:
```
12:14:54  docker : 
12:14:54  	grep: illegal option -- q
12:14:54  	Usage: grep -hblcnsviw pattern file . . .
12:14:54  	not on docker
```
still gets us the answer, but I guess the quiet option is not supported on whatever version of grep Solaris is using.  Approving as is, but we could also consider skipping this check on Solaris, as we know it would never be 'on docker'.

src/org/openj9/envInfo/MachineInfo.java
```diff
@@ -38,6 +38,7 @@ public class MachineInfo {
 	public static final String[] CPU_CORES_CMD_SOLARIS = new String[] {"bash", "-c", "psrinfo | wc -l"};
 	public static final String[] NUMA_CMD = new String[] {"bash", "-c", "numactl --show | grep 'No NUMA support available on this system"};
 	public static final String[] SYS_VIRT_CMD = new String[] {""};
+	public static final String[] CHECK_DOCKER_CMD = new String[] {"bash", "-c", "if grep -sq 'docker\\|lxc' /proc/1/cgroup; then echo 'running on Docker'; else echo 'not on docker'; fi"};
```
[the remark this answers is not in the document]
renfeiw: updated

renfeiw:
> Ran a few additional grinders to cover primaries [win](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5636/)/[mac](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5637/) and the platform where utilities is the most restricted in terms of options / [solaris](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5635/)).
> 
> Solaris hits this:
> 
> ```
> 12:14:54  docker : 
> 12:14:54  	grep: illegal option -- q
> 12:14:54  	Usage: grep -hblcnsviw pattern file . . .
> 12:14:54  	not on docker
> ```
> 
> still gets us the answer, but I guess the quiet option is not supported on whatever version of grep Solaris is using. Approving as is, but we could also consider skipping this check on Solaris, as we know it would never be 'on docker'.

Thanks for the note, we will update this check for Solaris later.

llxia (approved):
LGTM

src/org/openj9/envInfo/MachineInfo.java
```diff
@@ -38,6 +38,7 @@ public class MachineInfo {
 	public static final String[] CPU_CORES_CMD_SOLARIS = new String[] {"bash", "-c", "psrinfo | wc -l"};
 	public static final String[] NUMA_CMD = new String[] {"bash", "-c", "numactl --show | grep 'No NUMA support available on this system"};
 	public static final String[] SYS_VIRT_CMD = new String[] {""};
+	public static final String[] CHECK_DOCKER_CMD = new String[] {"bash", "-c", "if grep -sq 'docker\\|lxc' /proc/1/cgroup; then echo 'running on docker'; else echo 'not on docker'; fi"};
```
jerboaa, line 41: I'm not sure what this check is supposed to do, but If I'm reading this right it looks for `docker` or `lxc` in file `/proc/1/cgroup`? Is this supposed to be an or match? If so, isn't `-E` missing on the grep? Either way, this seems a strange thing to do. Why not use `.dockerenv`? See: https://github.com/moby/libnetwork/pull/815/commits/66ab744d19e534ed723fdb7b8df11a5e95f05630

While this would still not work if the container engine is `podman`, it would be a more "standard" approach. @smlambert thoughts?

smlambert: yep, the purpose of this check is to tell us where we are running... bare metal or in a container, happy to update it to a more standard approach

jerboaa: The thing is that it's not easily detectable for generic containers. If you care about `docker` only, using `.dockerenv` is the better thing to do. It makes it clear that this won't work generically, yet will return the right thing if run on `docker`. It won't work on podman or criu for example, though (but I don't think the current impl does either).

Issue #352, opened 2022-09-06
Enhance repeat feature in TKG


renfeiw:
Support two types of repetition in TKG:
- exit test when encountering a failure in last iteration
- keep running tests and collect failure rate in the result summary

Add fore_repeat option to allow repeat on AOT explicit test.

llxia:
Related: https://github.com/adoptium/TKG/pull/366

PR #356, opened 2022-09-13
Enhance TKG repeat feature

renfeiw:
- add ITERATIONS and AOT_ITERATIONS options
- add option to continue or stop after failure is detected

### DIFF
--- a/featureSettings.mk
+++ b/featureSettings.mk
@@ -12,16 +12,16 @@
 # limitations under the License.
 ##############################################################################
 
-ifneq (,$(findstring AOT,$(TEST_FLAG)))
-export TR_silentEnv=1
-export TR_Options=forceAOT
-export TR_OptionsAOT=forceAOT
-AOT_OPTIONS = -Xshareclasses:name=test_aot -Xscmx400M -Xscmaxaot256m
-ifndef TEST_ITERATIONS
-TEST_ITERATIONS = 2
-endif
+ifndef TKG_ITERATIONS
+	export TKG_ITERATIONS=1
 endif
 
-ifndef TEST_ITERATIONS
-TEST_ITERATIONS = 1
+ifneq (,$(findstring AOT,$(TEST_FLAG)))
+	export TR_silentEnv=1
+	export TR_Options=forceAOT
+	export TR_OptionsAOT=forceAOT
+	AOT_OPTIONS = -Xshareclasses:name=test_aot -Xscmx400M -Xscmaxaot256m
+	ifndef AOT_ITERATIONS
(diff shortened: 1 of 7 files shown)


renfeiw:
related: #352

llxia:
Could you update TKG_REPEAT and AOT_REPEAT to ITERATIONS and AOT_ITERATIONS in the description? Thanks

renfeiw:
> Could you update TKG_REPEAT and AOT_REPEAT to ITERATIONS and AOT_ITERATIONS in the description? Thanks

Thanks updated.

llxia (approved):
LGTM

smlambert (approved):
Thanks @renfeiw, this is nice!  I ran an open [Grinder](https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5703/) to see what the console output looks like.  

Note to self: We can likely revert/remove this [change](https://github.com/adoptium/aqa-tests/pull/1576/files) in Jenkinsfilebase given it is no longer needed to show iteration progress (and only works for Jenkins environment)

smlambert:
If I look at the console output, there is a change required in Jenkinsfilebase so that we do not multiply ITERATIONS by ITERATIONS, as I think we both run the make target ITERATIONS times, but also have a for loop in runTest() in Jenkinsfilebase, which we will need to remove or end up running ITERATIONS<sup>2</sup>.

Not sure if we want to merge this first or coordinate a PR to update Jenkinsfilebase and merge together.

renfeiw:
Yes, it would be better to merge together with this PR: https://github.com/adoptium/aqa-tests/pull/3989

renfeiw:
PR build: https://github.com/adoptium/aqa-tests/pull/3989

llxia:
hmm, we only have 1 TAP for the Grinder. I cannot tell we had 3 iterations in the TAP file https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5709/tapResults/

renfeiw:
> hmm, we only have 1 TAP for the Grinder. I cannot tell we had 3 iterations in the TAP file https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5709/tapResults/

This is the next step we need to add the iteration info into result summary.

renfeiw:
Just had a discussion with @llxia We think it's not the right stage to switch over completely to TKG iteration. So I will update the PR to use TKG_ITERATIONS and it will not conflict with Jenkins ITERATIONS.

renfeiw:
grinder: https://ci.adoptopenjdk.net/view/Test_grinder/job/Grinder/5712/

PR #367, opened 2022-11-11
Add failed testcase name in tap file

sophia-guo:
Related https://github.com/adoptium/aqa-tests/issues/4146

Signed-off-by: Sophia Guo <sophia.gwf@gmail.com>

### DIFF
--- a/scripts/resultsSum.pl
+++ b/scripts/resultsSum.pl
@@ -157,13 +157,9 @@ sub resultReporter {
 						$tapString .= "  ---\n";
 						if (($diagnostic eq 'failure') || ($diagnostic eq 'all')) {
 							if ($buildList =~ /openjdk/) {
-							    my @lines = split('\\n', $output);
 								my $failureTests = "";
-								foreach my $line (@lines)
-								{
-									if ( $line =~ /TEST: /) {
-										$failureTests .= $line . "\n";
-									}
+								while ( $output =~ /([-]{50}.)(TEST: )(.*?)(\.java|\.sh)/sg ) {
+									$failureTests .= "$2$3$4\n";
 								}
 								if ( $failureTests eq "" ) {
 									# Output of dump or other non-test failures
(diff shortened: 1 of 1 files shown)


sophia-guo:
With failed testcase: https://ci.adoptopenjdk.net/job/Grinder/6181/
With internal errors: https://ci.adoptopenjdk.net/job/Grinder/6180/
No errors on Windows: https://ci.adoptopenjdk.net/job/Grinder/6179/

karianna (approved):
LGTM

smlambert:
I like this, thanks @sophia-guo !  I will leave the 2nd / final review for @renfeiw (as Lan has explicitly asked for it).  

On the topic of adding additional information to the TAP files, separate from this work (and not expected to be covered by this PR), I was recently wondering if we could also include the line of jtreg output that summarized the total testcases in the target.  

`Test results: passed: 39; failed: 21`

scripts/resultsSum.pl
```diff
@@ -156,7 +156,25 @@ sub resultReporter {
 						$tapString .= "not ok " . $numOfTotal . " - " . $result . "\n";
 						$tapString .= "  ---\n";
 						if (($diagnostic eq 'failure') || ($diagnostic eq 'all')) {
-							$tapString .= $output;
+							if ($buildList =~ /openjdk/) {
+							    my @lines = split('\\n', $output);
+								my $failureTests = "";
+								foreach my $line (@lines)
+								{
+									if ( $line =~ /TEST: /) {
```
renfeiw, line 164: Is it possible to match the 50 hyphens before and then the `TEST:`
```
11:44:10  --------------------------------------------------
11:44:12  TEST: java/util/stream/test/org/openjdk/tests/java/util/stream/FilterOpTest.java
```
This could be more reliable compared to only `TEST:`. In an extreme case, if we have not only openjdk but also other folders in the buildlist, it could be more problematic.

sophia-guo: Updated. Trying with multiple line query doesn't work well ( works in my local environment and failed with builds). Decided just using the simple loop, which works and feels simple and straightforward.

sophia-guo:
Grinder in Eclipse temurin https://ci.eclipse.org/temurin-compliance/job/Grinder/2221/

renfeiw (approved):
LGTM

PR #377, opened 2022-11-25
Fix dependency actions and restrict permissions

tellison:
Fix dependency GitHub action to hash values and restrict to required permissions.
Part of https://github.com/adoptium/adoptium/issues/187


smlambert (approved):

karianna (approved):
LGTM

PR #381, opened 2022-12-07
Add jtreg 7.1.1+1 dependency

luhenry:


scripts/getDependencies.pl
```diff
@@ -150,6 +150,13 @@
 		shafn => 'jtreg_7_1.tar.gz.sha256sum.txt',
 		shaalg => '256'
 	},
+	jtreg_7_1_1_1 => {
+		url => 'https://ci.adoptopenjdk.net/job/dependency_pipeline/lastSuccessfulBuild/artifact/jtreg/jtreg-7-1-1+1.tar.gz',
+		fname => 'jtreg_7_1_1_1.tar.gz',
+		shaurl => 'https://ci.adoptopenjdk.net/job/dependency_pipeline/lastSuccessfulBuild/artifact/jtreg/jtreg-7-1-1+1.tar.gz.sha256sum.txt',
+		shafn => 'jtreg_7_1_1_1.tar.gz.sha256sum.txt',
+		shaalg => '256'
+	},
```
llxia, line 159: Thanks @luhenry . Can we remove jtreg_7_1?

karianna: We seem to keep multiple versions of the older mainlines - maybe we did that so we could reproduce?

karianna (approved):

scripts/getDependencies.pl
llxia: Not really. We use different versions of jtreg for different JDK versions. See https://github.com/adoptium/aqa-tests/blob/master/openjdk/build.xml#L29-L56

According to the code above, jtreg_7_1 is only used for JDK20+. If we update it to jtreg_7_1_1_1 for JDK20+, I think we should remove the old version jtreg_7_1.

luhenry: Let's remove it in two phases then? First merge this PR as-is, with the relevant update in aqa-tests, then the removal once it's been merged in aqa-tests.

PR #412, opened 2023-02-28
add the ability to pass EXTRA_DOCKER_ARGS via the command line

gdams:
This PR allows me to pass `DOCKER_ARGS` as a command line value:

e.g:

```bash
make _testList TESTLIST=tfb-spring-verify EXTRA_DOCKER_ARGS ="--cpus=1.0 --memory=8g"
```

to then be used in a playlist.xml file:

```xml
	<test>
		<testCaseName>tfb-spring-verify</testCaseName>
		<command>export DOCKER_ARGS=$(Q)$(EXTRA_DOCKER_ARGS)$(Q); bash $(Q)$(TEST_RESROOT)$(D)FrameworkBenchmarks/tfb$(Q) --mode verify --test spring; \
		$(TEST_STATUS)</command>
		<levels>
			<level>sanity</level>
		</levels>
		<groups>
			<group>perf</group>
		</groups>
	</test>
```

### DIFF
--- a/settings.mk
+++ b/settings.mk
@@ -380,7 +380,7 @@ setup_%: testEnvSetup
 	@$(ECHO) set JAVA_HOME to $(JAVA_HOME)
 	@$(ECHO) set SPEC to $(SPEC)
 	@$(ECHO) set TEST_FLAG to $(TEST_FLAG)
-	@$(ECHO) set DOCKER_ARGS to $(DOCKER_ARGS)
+	@$(ECHO) set EXTRA_DOCKER_ARGS to $(EXTRA_DOCKER_ARGS)
 
 	@$(MKTREE) $(Q)$(TESTOUTPUT)$(Q)
 	@$(ECHO) Running $(TESTTARGET) ...
--- a/settings.mk
+++ b/settings.mk
@@ -380,6 +380,7 @@ setup_%: testEnvSetup
 	@$(ECHO) set JAVA_HOME to $(JAVA_HOME)
 	@$(ECHO) set SPEC to $(SPEC)
 	@$(ECHO) set TEST_FLAG to $(TEST_FLAG)
+	@$(ECHO) set DOCKER_ARGS to $(DOCKER_ARGS)
 
 	@$(MKTREE) $(Q)$(TESTOUTPUT)$(Q)
 	@$(ECHO) Running $(TESTTARGET) ...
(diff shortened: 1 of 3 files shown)


smlambert:
Thanks @gdams .

Some notes:
- We try to align environment variables used by TKG with parameters used in the Jenkins groovy scripts when possible, and when they serve the 'same purpose'
- Currently we have a parameter called EXTRA_DOCKER_ARGS used for the external test group to pass in docker arguments
- we should look at this use case, where the test target itself is expecting to be run in an environment where docker is available (and is there a check that says it is not applicable to run in certain environments)

Trying to think about other use cases that will likely come up as well (other container based tests that are not currently a part of the external group of tests).
- dev.openjdk targets were recently added to house the container based testing from openjdk
- an outstanding PR from Alibaba to run some perf tests inside of containers 
- smoke tests that would best be run in containers (to avoid having extra build libs laying around and not having a clean environment to start from)

keithc-ca:
The example needs a small tweak; the value needs to be quoted to properly handle spaces:
```
export DOCKER_ARGS=$(Q)$(DOCKER_ARGS)$(Q); ...
```

gdams:
> The example needs a small tweak; the value needs to be quoted to properly handle spaces:

Updated - thanks for the pointer

renfeiw:
TKG is relying on make and based on make's behavior, anything you passed on the command line you can use it in `$()`. So we don't need to change anything in TKG, you should be able to use EXTRA_DOCKER_ARGS in test command.

gdams:
> TKG is relying on make and based on make's behavior, anything you passed on the command line you can use with `$()`. So we don't need to change anything in TKG, you should be able to use EXTRA_DOCKER_ARGS in playlist test command.

@renfeiw that didn’t work for me when I tried with the perf test playlist.xml file. Have you got a working example?

keithc-ca:
> TKG is relying on make and based on make's behavior, anything you passed on the command line you can use with `$()`. So we don't need to change anything in TKG, you should be able to use EXTRA_DOCKER_ARGS in playlist test command.

Yes, make can reference `$(EXTRA_DOCKER_ARGS)`, but only if `EXTRA_DOCKER_ARGS` is in the environment or passed explicitly on the command-line (and not overridden by a macro definition).

gdams:
Oh yes, for some reason this seems to be working now without the changes - closing

PR #419, opened 2023-03-11
Preserve all internal comments

### LINKED ISSUE #395: Improve remove_duplicate_tests.py
With https://github.com/adoptium/TKG/pull/393, we have remove_duplicate_tests.py to remove duplicate tests. However, there are some enhancements that we can make:

- [ ] find a way to persevere comments (whether the license or actual comments within the root element). For example https://stackoverflow.com/questions/33573807/faithfully-preserve-comments-in-parsed-xml. The license can be different from repo to repo. Users should not need to provide `--license-file`. Also, actual comments in the playlist.xml should be preserved.
- [ ] We should compare the content instead of simply removing elements with the duplicated `<testCaseName>`. Only remove if everything under `<test>` is the same. If not, we should print a warning message. 
- [ ] Update remove_duplicate_tests.py to take a git repo. remove_duplicate_tests.py should git clone the repo and check all subdirectories to find playlist.xml and remove duplicate tests if possible.
- [ ] create a GitHub workflow for this task. On a schedule (i.e., biweekly), run remove_duplicate_tests.py against all repos and create auto PR against specific repo if needed.
https://github.com/adoptium/aqa-tests
https://github.com/eclipse-openj9/openj9
https://github.com/adoptium/temurin-build

ArkaprabhaChakraborty:
While the external comments outside the root element tag cannot be preserved due to how the XML Tree is built by the TreeBuilder, the internal comments can be saved and this PR aims at that task

Related #395

### DIFF
--- a/scripts/remove_duplicate_tests.py
+++ b/scripts/remove_duplicate_tests.py
@@ -21,12 +21,12 @@ def parse_apache_2_license(license_file):
 
         license_lines = license_lines[190:201]
         f.close()
-    # add the ## line to beginning of each line
-    license_lines = ''.join(['# ' + line.lstrip() for line in license_lines])
+    license_lines = ''.join(['#' + line for line in license_lines])
     return license_lines
 
 def remove_duplicate_test_tags(xml_file, license_file):
-    tree = ET.parse(xml_file)
+    tree = ET.parse(xml_file, 
+                    parser=ET.XMLParser(target=ET.TreeBuilder(insert_comments=True)))
     root = tree.getroot()
     test_case_names = []
     for test in root.findall('test'):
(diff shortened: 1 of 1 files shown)


karianna (approved):

karianna:
@ArkaprabhaChakraborty Are you able to update the PR?

ArkaprabhaChakraborty:
Yep, previously if comments were within the root element's body they were not preserved. So if we consider the following input:
```xml
<playlist xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../TKG/playlist.xsd">
     <!--Test playlist for the OpenJ9 project.-->
          <test>
		<testCaseName>jcstress_BooleanFencedTest</testCaseName>
		<command>$(JAVA_COMMAND) $(JVM_OPTIONS) -jar $(Q)$(LIB_DIR)$(D)jcstress-tests-all-20220908.jar$(Q) $(APPLICATION_OPTIONS) -t BooleanFencedTest; \
			$(TEST_STATUS)</command>
		<levels>
````
So we would have lost the following tag:
```xml
 <!--Test playlist for the OpenJ9 project.-->
```

But now it will preserve all the comments under the root element. So the output now will be the same as the above input

llxia (approved):

PR #490, opened 2024-01-23
Use cut instead of $CUT

RadekCap:
Add check of CUT.

CUT is needed for a proper setting of LIB_DIR in CYGWIN.

### DIFF
--- a/makefile
+++ b/makefile
@@ -25,10 +25,6 @@ export JAVA_HOME := $(TEST_JDK_HOME)
 $(info JAVA_HOME is set to $(JAVA_HOME))
 endif
 
-ifndef CUT
-$(error Please provide CUT value.)
-endif
-
 D = /
 MKTREE = mkdir -p
 SUBDIRS = ..
@@ -41,7 +37,7 @@ ifndef LIB_DIR
 endif
 
 UNAME := uname
-UNAME_OS := $(shell $(UNAME) -s | $(CUT) -f1 -d_)
+UNAME_OS := $(shell $(UNAME) -s | cut -f1 -d_)
 ifeq ($(findstring CYGWIN,$(UNAME_OS)), CYGWIN)
 	LIB_DIR:=$(shell cygpath -w $(LIB_DIR))
 endif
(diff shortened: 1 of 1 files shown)


sophia-guo:
cut is only use once for CYGWIN check [UNAME_OS := $(shell $(UNAME) -s | $(CUT) -f1 -d_)](https://github.com/adoptium/TKG/blob/master/makefile#L43). May just use `cut` instead.

judovana:
@sophia-guo  can you  please elaborate more?  Do yo recomend to use `cut` instead `$(CUT)` ? Whats the benefit? The cut will still be missing from path. And it will enad as `command not found`...

I'm missing some detection of the cut. Where do you expect it to be set? Can  you point us?

judovana:
@RadekCap  maybe you can add fallback?

if `$(CUT)` is missing, maybe yo may check path for `cut` it, and if it is there, then use that, if not then indeed die...

sophia-guo:
I mean use `cut` directly there is no need to define `CUT` as it is used only once and presumably that will be true for a long time. 

@RadekCap and I have talked about it in a meeting and @RadekCap will update the PR accordingly.

RadekCap:
@sophia-guo updated as requested.

I've added a debug message for LIB_DIR as well.

judovana:
I think the verification that the cut is on path should be  done. Also i think that calling cut without $(CUT) is bad code.

sophia-guo:
I think the `cut` command comes pre-installed in most Linux distributions. Though agreed though @judovana it doesn't hurt to verify.

karianna (approved):

sophia-guo (approved):

judovana:
```
-UNAME_OS := $(shell $(UNAME) -s | $(CUT) -f1 -d_)
+UNAME_OS := $(shell $(UNAME) -s | cut -f1 -d_)
```

Just to summ this up - we agreed it should be checked - but  the patch was approvred without the check.
In addition we caused (microscopical) regression, that if I have CUT declared, it will be ignored.

Am I right or am reading it wrong? 

I know the proper gues/set/check if is nto exactly plesant, but Radek's original commit was very close to it.  Also I agree it is super minor thing, but I feel very very uncomfortable about it:( Do we agree on finish it properly (@RadekCap  please?), or do you really want to set this precedent?

PR #493, opened 2024-01-31
Add hamcrest-core.jar and bcprov-jdk18on.jar in getDependencies

llxia:
related: https://github.com/adoptium/aqa-tests/issues/4979 and https://github.com/adoptium/aqa-tests/pull/5024

PR #494, opened 2024-02-05
Support vendor names with spaces in compile.mk

pushkarnk:
https://github.com/adoptium/TKG/pull/493 added the JDK_VENDOR property to the compile command. This value is picked up from the "java.vendor" property of the test JDK in TEST_JDK_HOME. If the "java.vendor" value has a whitespace, it breaks the new compile command. Hence, the commit proposes using "" around $(JDK_VENDOR) in COMPILE_CMD.

The new TKG caused failure on Ubuntu because the Ubuntu JDK builds have "java.vendor" set to "Private Build".


pushkarnk:
@llxia Can you review this please?

llxia:
vendor is a key value that we will use to enable/disable tests. Each vendor would have its own value https://github.com/adoptium/TKG/blob/6ab3f8f43bba41f8e81f0a041f48931f8dd15001/src/org/openj9/envInfo/JavaInfo.java#L167-L198

What is the full output of your "java.vendor"? Just "Private Build" or it contains vendor name (i.e., Canonical Private Build)?

By setting "java.vendor" to just "Private Build" (a generic name), it would be hard to enable/disable tests based on the specific vendor build (i.e., Canonical build) in the future. 

Ideally, the correct way is to add "java.vendor" support above [JavaInfo.java](https://github.com/adoptium/TKG/blob/6ab3f8f43bba41f8e81f0a041f48931f8dd15001/src/org/openj9/envInfo/JavaInfo.java#L167-L198):
```
else if (vendorLC.contains("canonical")) { 
            return "canonical"; 
}
```

pushkarnk:
Got your point @llxia. Thank you.

PR #519, opened 2024-03-14
Update GH permissions

annaibm:
Add GH permissions
- Add GH action permissions for testTKG job

### DIFF
--- a/.github/workflows/testTKG.yml
+++ b/.github/workflows/testTKG.yml
@@ -4,6 +4,10 @@ on:
     types: [created]
 jobs:
   testTKG:
+    permissions:
+        contents: read
+        issues: write
+        pull-requests: write
     runs-on: ubuntu-latest
     if: startsWith(github.event.comment.body, 'run tkg-test')
     steps:


annaibm:
run tkg-test

annaibm:
run tkg-test

annaibm:
run tkg-test

.github/workflows/testTKG.yml
```diff
@@ -4,10 +4,11 @@ on:
     types: [created]
 jobs:
   testTKG:
+    permissions: write-all
```
smlambert, line 7: Let's not turn on write permissions for everything, as we only need it for a few things from the list https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#defining-access-for-the-github_token-scopes 

Also, I do not think you can test a Github actions change by using the workflow itself, we will likely just have to make a precise change, merge it, then check if it is behaving as we want (otherwise test it in a separate personal repo).

annaibm: Tested within the personal forked repo Actions and its working fine. So likely a permission issue when PR from forked repository is tested from main repo.

smlambert (approved):

llxia (approved):

Issue #510, opened 2024-03-08
Enhance Test result summary with testcases results


sophia-guo:
Currently test result summary is provided by Targets level, which doesn't provide an accurate test information per test cases. It might be helpful if test summary per test cases could be provided. 

Both of JCK and JDK tests provide test summary per testcase (`Test results: passed: 41; failed: 5; error:1; skipped: 1`), which means this can be done for JCK and JDK tests.

Console output for each run could be something like:

```
TEST TARGETS SUMMARY
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++:
DISABLED test targets:
	jdk_awt_1
        ...
	jdk_sound_0

PASSED test targets:
        jdk_beans_2  - Test results: passed: 41; skipped: 1
        ...
  	jdk_other_2 - Test results: passed: 21

FAILED test targets:
        jdk_tools_2 - Test results: passed: 41; failed: 5; error:1; skipped: 1
        ...
        jdk_lang_1 -  Test results: passed: 541; failed: 5
TOTAL: 40   EXECUTED: 7   PASSED: 6   FAILED: 1   DISABLED: 11   SKIPPED: 22
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++:
 
TESTCASES RESULTS SUMMARY: passed: 214; failed: 1; error: 0; skipped: 0  
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```
Correspondingly tap file could also include those information for passed ones if needed ( failed one already include those information)

```
#SHA.txt content: 
...
# TEST TARGETS RESULTS SUMMARY: TOTAL: 2   EXECUTED: 2   PASSED: 1   FAILED: 1   DISABLED: 0   SKIPPED: 0
# TESTCASES RESULTS SUMMARY: passed: 214; failed: 1; error: 0; skipped: 0 
1..2
ok 1 - jdk_math_0
  ---
    output:
      |
        Test results: passed: 39
    duration_ms: 100014
  ...
not ok 2 - jdk_security1_0
  ---
    output:
      |
        Failed test cases: 
        TEST: java/security/cert/CertPathBuilder/akiExt/AKISerialNumber.java
        Test results: passed: 175; failed: 1
    duration_ms: 95974
  ...
```

Thoughts @smlambert @jiekang ?

sophia-guo:
Related https://github.com/adoptium/aqa-tests/issues/5121

[...]

PR #520, opened 2024-03-14
Limit overall test cases summary for all targets to openjdk and jck only

sophia-guo:
Some of other category tests might also run with jtreg. In that case testcase summary per target is still useful.

Related #510

### DIFF
--- a/scripts/resultsSum.pl
+++ b/scripts/resultsSum.pl
@@ -317,7 +317,7 @@ sub resultReporter {
 	print "$testTargetStatus\n";
 	
 	print "++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++\n\n";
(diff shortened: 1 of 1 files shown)


sophia-guo:
jck and openjdk keep original behavior:
temurin-compliance/job/Grinder/4107/console
https://ci.adoptium.net/view/Test_grinder/job/Grinder/9147/

functional without jtreg tests - no testcase results summary
https://ci.adoptium.net/view/Test_grinder/job/Grinder/9153/console
functional with jtreg tests - no overrall testcases results summary, testcases results summary per targets is enabled
https://ci.adoptium.net/view/Test_grinder/job/Grinder/9152/console

smlambert (approved):

karianna (approved):

PR #528, opened 2024-04-12
Disable Mode158 on AArch64 Linux

knn-k:
This commit disables Mode158 on AArch64 Linux.
Mode158 is described as "test 3bit shifting(z/OS only)".


knn-k:
Definition of Mode158: https://github.com/adoptium/TKG/blob/232682673cdbd1a4b818b2b7e13bf4ea39f844c3/resources/modes.xml#L1139-L1140

karianna (approved):

llxia (approved):

PR #542, opened 2024-05-02
Use jtreg 7.3.1+1 for JDK11 tests

llxia:
fixes: https://github.com/eclipse-openj9/openj9/issues/19435

### DIFF
--- a/scripts/getDependencies.xml
+++ b/scripts/getDependencies.xml
@@ -29,22 +29,8 @@
 			<then>
 				<property name="jtregTar" value="jtreg_5_1_b01"/>
 			</then>
-			<elseif>
-				<!-- version 11 -->
-				<matches pattern="^11$" string="${JDK_VERSION}"/>
-				<then>
-					<property name="jtregTar" value="jtreg_6_1"/>
-				</then>
-			</elseif>
-			<elseif>
-				<!-- versions 17, 21, 22, 23 -->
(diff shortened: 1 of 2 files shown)


llxia:
https://openj9-jenkins.osuosl.org/job/Grinder/3561/

scripts/getDependencies.xml
```diff
@@ -30,15 +30,8 @@
...
-				<matches pattern="^11$" string="${JDK_VERSION}"/>
-				<then>
-					<property name="jtregTar" value="jtreg_6_1"/>
-				</then>
-			</elseif>
-			<elseif>
-				<!-- versions 17, 21, 22, 23 -->
-				<matches pattern="^(17|2[123])$" string="${JDK_VERSION}"/>
+				<!-- versions 11, 17, 21, 22, 23 -->
+				<matches pattern="^(11|17|2[123])$" string="${JDK_VERSION}"/>
```
pshipton, line 34: Why do we need to check versions and fail, requiring an update every time there is a new jdk version? Could it just load 7.3.1.1 until we find a newer jdk version requires something different? The OpenJDK testing checks the version and fails if it's not recent enough.

llxia: I see no need to check the version either. I will update the code

llxia:
https://openj9-jenkins.osuosl.org/job/Grinder/3564/

smlambert (approved):

karianna (approved):

Issue #535, opened 2024-04-18
Add vendor Temurin in TKG


llxia:
Notice the following in the Grinder: https://ci.adoptium.net/view/Test_grinder/job/Grinder/9702/console

```
14:23:20  System.getProperty('java.vendor')=Temurin
14:23:20  
14:23:20  Warning: cannot determine vendor, use System.getProperty('java.vendor')=Temurin directly.
14:23:20  
```

code: https://github.com/adoptium/TKG/blob/a9b6c3e1ef19b948a47cc6e44484ebf40aa9bd78/src/org/openj9/envInfo/JavaInfo.java#L167-L200

Hajer-Dahmani:
I will take this.

PR #554, opened 2024-05-26
added vendor temurin in TKG

Hajer-Dahmani:
added vendor temurin in TKG

Fixes: #535

signed off by Hajer Dahmani <hdahmani@andrew.cmu.edu>


karianna:
Not sure if eclipse and temurin should be treated as the same thing here

smlambert:
> Not sure if eclipse and temurin should be treated as the same thing here

This PR handles cases when System.getProperty('java.vendor') returns Temurin.  While Temurin is not actually a vendor, we need to handle the case when JDK builds actually return this value.

Issue #640, opened 2024-11-26
JVM_OPTIONS="" as an environment variable will be considered as set and variations be ignored.


sophia-guo:
In some environment JVM_OPTIONS="" but still works as an environment variable and variations are ignored mistakenly. 

For example in jenkins with dockerimage all build parameters will be considerred as environment variable even if it's empty string.
https://ci.adoptium.net/job/Grinder/11686/consoleFull . The condition should enhance to consider the empty string.

```
[Pipeline] sh
16:48:20  + docker pull ghcr.io/adoptium/test-containers:ubuntu2204
16:48:20  ubuntu2204: Pulling from adoptium/test-containers
16:48:20  Digest: sha256:1b1e379049f788592687c3bb5d268c0827f1fc15226ffe4188c885f1f72fea20
16:48:20  Status: Image is up to date for ghcr.io/adoptium/test-containers:ubuntu2204
16:48:20  ghcr.io/adoptium/test-containers:ubuntu2204
...
16:48:28  + printenv
16:48:28  JENKINS_HOME=/home/jenkins/.jenkins
16:48:28  OPENJ9_REPO=https://github.com/eclipse/openj9.git
16:48:28  DOCKER_REGISTRY_DIR=
16:48:28  USE_TESTENV_PROPERTIES=false
16:48:28  BUILD_LIST=openjdk
16:48:28  VENDOR_TEST_DIRS=
16:48:28  CI=true
16:48:28  OPENJ9_BRANCH=master
16:48:28  SDK_RESOURCE=nightly
16:48:28  HOSTNAME=027c99e76dd4
16:48:28  RUN_CHANGES_DISPLAY_URL=https://ci.adoptium.net/job/Grinder/11686/display/redirect?page=changes
16:48:28  ARTIFACTORY_ROOT_DIR=
16:48:28  ADOPTOPENJDK_REPO=https://github.com/gdams/openjdk-tests.git
16:48:28  UPSTREAM_JOB_NUMBER=
16:48:28  JVM_OPTIONS=
16:48:28  NODE_LABELS=ubuntu x86-64 x64 ci.agent.dynamic sw.os.linux test-linux-x64-8d33b3 hw.arch.x86 ubuntu2204
16:48:28  VENDOR_TEST_BRANCHES=
16:48:28  HUDSON_URL=https://ci.adoptium.net/
16:48:28  TARGET=extended.openjdk
[...]

PR #642, opened 2024-11-27
Enhance the check of envTf variable

sophia-guo:
Similar to #640 with jenkins docker pull images, envTf could be an empty string.


smlambert (approved):

llxia (approved):
LGTM

PR #668, opened 2025-01-30
get sha for external tests

sophia-guo:
Fix https://github.com/adoptium/aqa-tests/issues/5825

### DIFF
--- a/scripts/getSHAs.sh
+++ b/scripts/getSHAs.sh
@@ -22,7 +22,6 @@ usage ()
 	echo 'Usage : '
 	echo '                --test_root_dir: optional'
 	echo '                --shas_file: optional, the file to write the sha info to. Default is to ../SHAs.txt'
-
 }
 
 parseCommandLineArgs()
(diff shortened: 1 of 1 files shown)


sophia-guo:
https://ci.adoptium.net/view/Test_grinder/job/Grinder/12666/artifact/aqa-tests/TKG/output_17383599695618/Grinder_20250131214432.tap/*view*/

sophia-guo:
@smlambert

scripts/getSHAs.sh
```diff
@@ -62,14 +61,43 @@ getSHAs()
...
+					github_url=$(grep '^github_url=' "$propertiesFile" | cut -d"=" -f2 | tr -d '"')
+					tag_version=$(grep '^tag_version=' "$propertiesFile" | cut -d"=" -f2 | tr -d '"')
+
+					if [[ -z "$github_url" || -z "$tag_version" ]]; then
+						echo "No github_url or tag_version in $propertiesFile"
+						continue
+					fi
+
+					# Retrieve the SHA using github_url and tag_version
+					sha=$(git ls-remote "$github_url" "refs/tags/$tag_version" | awk '{print $1}')
```
smlambert, line 92: Thanks @sophia-guo, is there ever a scenario where git ls-remote call will fail and if so, what is the fallback plan?

sophia-guo: Updated. Only do the tracking if sha is not empty, which scenario means the test will fail right at the beginning stage of setting up the environment and there is not any test output at all. Hence no need to track the sha.

smlambert (approved):

karianna (approved):

PR #685, opened 2025-03-20
Correct post test Perl resultsSum processing to correctly handle commas and Test Results line

andrew-m-leonard:
Fixes https://github.com/adoptium/aqa-tests/issues/6081

Improve resultsSum.pl to recognise openjdk or jck Test results lines.
Some of the regex's were not allowing for ",", and were unnecessarily trying to group the (numeric) value, when just recoginising the error|failed is all that was needed.

Test builds:
TC/job/Test_openjdk24_hs_extended.jck_aarch64_mac/22/console - SUCCESSFUL test result summary and tap file content, and failed tests ReRun
extended.openjdk test: https://ci.adoptium.net/view/Test_openjdk/job/Test_openjdk21_hs_extended.openjdk_aarch64_linux/133/ - SUCCESSFUL test result summary and tap file content

### DIFF
--- a/scripts/resultsSum.pl
+++ b/scripts/resultsSum.pl
@@ -148,7 +148,8 @@ sub resultReporter {
 						$endTime = $1;
 						$tapString .= "    duration_ms: " . ($endTime - $startTime) . "\n  ...\n";
 						last;
-					} elsif ($result =~  /(Test results: .*)(passed|skipped|failed|error)(: .*\d{1,}$)/) {
+					} elsif ($result =~ /(Test results: .*)(passed|skipped|failed|error)/) {
+						# We have the Test results line
 						$testCasesPerTargetSummary = $result;
 						chomp($testCasesPerTargetSummary);
 						push (@testCasesResults, $result);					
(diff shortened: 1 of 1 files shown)


scripts/resultsSum.pl
```diff
@@ -148,7 +148,8 @@ sub resultReporter {
 						$endTime = $1;
 						$tapString .= "    duration_ms: " . ($endTime - $startTime) . "\n  ...\n";
 						last;
-					} elsif ($result =~  /(Test results: .*)(passed|skipped|failed|error)(: .*\d{1,}$)/) {
```
sophia-guo, line 151: This works as console output. as `Test results: passed: 4; failed: 6; skipped: 3,912` does show in the TEST TARGETS SUMMARY part. So the easiest way is to update \d{1,}$)/ to .*\d{1,}$)/ in line 204 and 218, which will match coma part.

andrew-m-leonard: yeah I didn't like the way the ".*" was being used to match the ",", I think that was by accident. The group is giving the "illusion" it is capturing the "number", when it is not...
Hence I took the decision to simplify it and remove the "iilusion", as there is no need to capture the number here, and I think my suggestion simplifies the understanding of what it is doing?
It's illusions like this that imo lead to misstakes.

Thoughts?

andrew-m-leonard: i'm happy either way.... i just prefer simplicity!

andrew-m-leonard: Also the reason line :218 incorrectly contains "skipped" is so that it matches, because skipped is the last keyword, so previously the following will match as failed:
Test results: passed: 4; skipped: 3
but removing skipped from the old version would not then match the following:
Test results: passed: 4; failed: 6; skipped: 3

sophia-guo: Agree, since no need to capture the number here this way is simpler.

sophia-guo (approved):

smlambert (approved):

PR #688, opened 2025-04-01
Made JTREG_JAR declared once

judovana:
The path to jtreg.jar is decalred on several places in aqa-tests. This PR is declaring it once, and is allowing it tto be reused. It is honouring the user declaration, if any

### DIFF
--- a/settings.mk
+++ b/settings.mk
@@ -180,6 +180,14 @@ endif
 
 TESTNG=$(LIB_DIR)$(D)testng.jar$(P)$(LIB_DIR)$(D)jcommander.jar
 RESOURCES_DIR=$(JVM_TEST_ROOT)$(D)TestConfig$(D)resources
+ifdef JTREG_JAR
+    $(ECHO) "Custom jtreg jar declared: $(JTREG_JAR)"
+   CUSTOM_JTREG_JAR=true
+else
+   CUSTOM_JTREG_JAR=false
+   JTREG_JAR=$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar
+endif
+
 
 #######################################
 # cmdlinetester jars


smlambert:
We need a clear description of the problem that this PR is trying to resolve so we can suggest the best way to solve the problem.  If we are to support swapping in a custom jtreg.jar, we need to consider how to track and report what is being installed and used in test pipelines.  

In the feature request, listing the details of where jtreg.jar is pulled in and declared will be necessary.  I am only aware of it being pulled in as a testDependency via TKG, and also somewhat silently from the rh-openjdk wrapper scripts, which is one of the reasons using those wrapper scripts is not preferred.

judovana:
Hi! Thanx for reply!

> We need a clear description of the problem that this PR is trying to resolve so we can suggest the best way to solve 

We failed to upstream the fast/slow debug extension point to jtregs: https://github.com/openjdk/jtreg/pull/235#issuecomment-2713065800 

It will take years (if ever) before I or @jandrlik/@andrlos will get back to it with second attempt.

If the upstream jtreg jar will be missing the crash detection and propagation for next few years, and I want to use aqa-test for slow/fastdebug testing, then I need custom jtreg.jar.

Currently I have to sed all that lines as in https://github.com/adoptium/aqa-tests/pull/6125

> the problem. If we are to support swapping in a custom jtreg.jar, we need to consider how to track and report what is being installed and used in test pipelines.
> 

I would love to do that. And already before this PR, i was trying to look where and how.  But all the dependence checks seems to run before - including jtreg version selector.
The only idea I had was t run some SHA and java -jar jtreg.jar --version  in/after that  ifdef  in this PR.
I dropped it, as I come to conclusion, that actually nothing is touching  that, and the only touch would be strictly intentional.  

If you have any pointers where to look or what to enhance, would be awesome.

One internal view to myslef - I believe that any dependence should be replaceable without hacking and together with that, I'm really happy how aqa-tests/tkg and comp. are keeping the versions clear and pure.  So you can see, I do not have it easy with myself...

> In the feature request, listing the details of where jtreg.jar is pulled in and declared will be necessary. I am only aware of it being pulled in as a testDependency via TKG, and 

But that was not broken, right?

> also somewhat silently from the rh-openjdk wrapper scripts, which is one of the reasons using those wrapper scripts is not preferred.

Please, no. After the chat we had, I had reworked it all and it is strictly honouring what aqa-tests gives. Only I kept the ability to overwrite it if it is necessary, exactly for cases like this one.  I have definitely understood and signed under   purity of clear dependences fully in aqa-tests/tkg control, and am not going to intentionally break that.

That also remained me, that some parts are using something like `"$(TEST_RESROOT)$(D)jtreg"`   That should probably be handled in this PR too.

judovana:
> That also remained me, that some parts are using something like "$(TEST_RESROOT)$(D)jtreg" That should probably be handled in this PR too.

Yah, that support needs small rework of this PR.  but afaik it is mandatory (if it ever can go in somehow)

settings.mk
```diff
@@ -180,6 +180,14 @@ endif
...
 TESTNG=$(LIB_DIR)$(D)testng.jar$(P)$(LIB_DIR)$(D)jcommander.jar
 RESOURCES_DIR=$(JVM_TEST_ROOT)$(D)TestConfig$(D)resources
+ifdef JTREG_JAR
+    $(ECHO) "Custom jtreg jar declared: $(JTREG_JAR)"
+   CUSTOM_JTREG_JAR=true
+else
+   CUSTOM_JTREG_JAR=false
+   JTREG_JAR=$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar
+endif
+
```
smlambert, line 190:
```suggestion
ifdef CUSTOM_JTREG_JAR
    $(ECHO) "Custom jtreg jar declared: $(CUSTOM_JTREG_JAR)"
   JTREG_JAR=$(CUSTOM_JTREG_JAR)
   # IS_CUSTOM_JTREG=true
else
   JTREG_JAR=$(TEST_RESROOT)$(D)jtreg$(D)lib$(D)jtreg.jar
   # IS_CUSTOM_JTREG=false
endif

```

I think the variable naming you chose will confuse people, so suggesting the above changes, since there _can_ always be a JTREG_JAR, but in some cases it will be overridden with a custom value. 

I would expect CUSTOM_JTREG_JAR to contain the path to a custom jtreg jar. If it is still needed, I would expect a boolean variable to be named something like IS_CUSTOM_JTREG or OVERRIDE_WITH_CUSTOM_JTREG.  Not actually sure you need it (maybe if later you want to query if a custom jtreg jar is in use).

smlambert (commented):
Adding some review comments now, is not a full review at this time, since I also want to consider whether we should also do as we do for pulling in official releases of testDependencies, verify the checksum / track SHAs etc.

settings.mk
judovana: Sounds reasoanble. the IS_CUSTOM_JTREG will become obsoleted as it will be easy to check  CUSTOM_JTREG_JAR. I will include it to rework due to missing `$(TEST_RESROOT)$(D)jtreg`

PR #698, opened 2025-04-28
Preserve encoding and tags while copying files on z/OS

psoujany:
This PR fixes https://github.ibm.com/runtimes/openj9-openjdk-jdk21-zos/pull/713#issuecomment-111565653 and https://github.ibm.com/runtimes/openj9-openjdk-jdk21-zos/issues/720

PR #704, opened 2025-05-19
Copy TKG compiled repo to BUILD_ROOT

psoujany:
This PR adds filename while copying TKG material from aqa-tests to BUILD_ROOT(jvmtest).


llxia:
Please explain the reason for this PR. What was broken? If you want to do the same thing as the original code, I would expect to copy *.mk as well.

```
<copy todir="${ascii.test.dir}">
					<fileset dir="${ascii.workspace.dir}">
						<include name="TKG/**" />
						<include name="*.mk" />
					</fileset>
				</copy>
```

llxia:
How did it pass in https://github.com/adoptium/TKG/pull/698#issuecomment-2882041377?

psoujany:
functional, system, openjdk tests aren't using TKG scripts once the test execution starts, hence they passed. MachineInfo targets failed by which we identified this.

llxia:
MachineInfo target is not in functional and openjdk, but it belongs to the system test. Also, MachineInfo target should be removed in the future.

psoujany:
mk files outside TKG doesn't need to be copied to jvmtest, mk files inside TKG are already been copied as part of     `<include name="TKG/**" />`.

Verification on non-z/OS
Grinder_CR/38625/

scripts/build_test.xml
```diff
@@ -102,15 +102,14 @@
...
-					<arg value="${ascii.workspace.dir}" />
+					<arg value="${ascii.workspace.dir}/TKG" />
 					<arg value="${ascii.test.dir}" />
 				</exec>
 			</then>
 			<else>
 				<copy todir="${ascii.test.dir}">
 					<fileset dir="${ascii.workspace.dir}">
 						<include name="TKG/**" />
-						<include name="*.mk" />
```
LongyuZhang, line 113: Discussed with @smlambert and @llxia , we still need this for local run.
@psoujany Could you update to add this line back? Thanks.

psoujany: Thank you @LongyuZhang , updated as suggested.

PR #755, opened 2025-10-15
Also download jtreg-8.1+1 and use for Java 26+

keithc-ca:
Inline test for Java 11 or 17 z/OS.

Depends on https://github.com/adoptium/ci-jenkins-pipelines/pull/1289.

### DIFF
--- a/scripts/getDependencies.xml
+++ b/scripts/getDependencies.xml
@@ -21,12 +21,6 @@
 	<!-- set default LIB property to all -->
 	<property name="LIB" value="all"/>
 	<target name="getJtregVersion">
-		<condition property="jtregOnZ">
-			<and>
-				<contains string="${SPEC}" substring="zos"/>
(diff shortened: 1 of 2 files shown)


scripts/getDependencies.xml
```diff
@@ -66,7 +63,7 @@
 			</elseif>
 			<else>
 				<!-- versions 26+ -->
-				<property name="jtregTar" value="jtreg_8_2"/>
+				<property name="jtregTar" value="jtreg_8_1_1"/>
```
karianna: Would we want to use the latest that is feasible?

karianna: Ah, OK, it looked like 8.1.1 vs 8.2

karianna (approved):

smlambert:
Now available https://ci.adoptium.net/view/Dependencies/job/dependency_pipeline/1707/artifact/jtreg/

PR #757, opened 2025-10-24
Use jtreg 8+2 for Java 25

keithc-ca:
Necessitated by openjdk change:
* [8361950: Update to use jtreg 8](https://github.com/openjdk/jdk25u/commit/9df689f143e43d92bb5d281985ce9c5c514f8803)

### DIFF
--- a/scripts/getDependencies.xml
+++ b/scripts/getDependencies.xml
@@ -58,7 +58,7 @@
 				<!-- version 25 -->
 				<matches pattern="^25$" string="${JDK_VERSION}"/>
 				<then>
-					<property name="jtregTar" value="jtreg_7_5_1_1"/>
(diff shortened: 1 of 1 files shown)


karianna (approved):

llxia:
Does it mean that we can remove jtreg_7_5_1_1: https://github.com/adoptium/TKG/blob/master/scripts/getDependencies.pl#L177-L183?

keithc-ca:
I think all of these jtreg versions will be unused by current Java versions:
* 6+1
* 6.1+1
* 7+1
* 7.1.1+1
* 7.2+1
* 7.3+1
* 7.5.1+1
* 7.5.2+1

If we want to prune them, I think that should be done separately.

PR #768, opened 2025-11-04
Added jtreg_7_5_1_1 to getDependency files for JDK25

Amrutha-Kanhirathingal:
jdk_security1 - java/security/SignedJar/spi-calendar-provider/TestSPISigned - NoSuchFileException

- Replaced jtreg_8_2 with jtreg_7_5_1_1 in getDependecny.xml

- Added jtreg_7_5_1_1 base jar info to getDependency.pl

Related: https://github.com/adoptium/aqa-tests/issues/6708

### DIFF
--- a/scripts/getDependencies.pl
+++ b/scripts/getDependencies.pl
@@ -174,6 +174,13 @@ my %base = (
 		shafn => 'jtreg_7_4_1.tar.gz.sha256sum.txt',
 		shaalg => '256'
 	},
+	jtreg_7_5_1_1 => {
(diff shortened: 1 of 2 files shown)


LongyuZhang (approved):

pshipton:
> Why? This will break testing of jdk25.

It will fix the OpenJ9 testing, because the change to use jtreg_8_2 is only in the master branch, but OpenJ9 uses the openj9 branch to which only OpenJDK builds are merged. Right now it's 25.0.1+8, next it will be 25.0.2. These are delivered to open source from Oracle, and may or may not include the jtreg_8_2 change. Not until development for 25.0.3 might open source changes in master be included.

PR #771, opened 2025-11-07
introducing test_java_home to allow for jre testing

andrlos:
related to the https://github.com/adoptium/aqa-tests/issues/6653

JAVA_TO_TEST does not have a counterpart based in the JDK installation in case USE_JRE is set to true
the idea is to have two counterparts:
TEST_JDK_HOME - always point to the JDK installation directory
TEST_JAVA_HOME - always points to the directory of the java that is being tested

second iteration would be to ditch JAVA_TO_TEST variable use in TCK in favor of TEST_JAVA_HOME variable

### DIFF
--- a/settings.mk
+++ b/settings.mk
@@ -169,10 +169,11 @@ ifdef JRE_IMAGE
    JRE_COMMAND:=$(Q)$(JRE_IMAGE)$(D)bin$(D)java$(Q)
 endif
 
-JAVA_TO_TEST = $(JAVA_COMMAND)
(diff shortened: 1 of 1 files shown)


andrlos:
okay, I was under the impression that we do not want to retain JAVA_TO_TEST.. let me circle back to the original discussion then :D

smlambert:
I just meant we no longer use the USE_JRE switch for TC testing, which was the original reason it was introduced.  Now that JAVA_TO_TEST is added, it is not going away.

Issue #770, opened 2025-11-06
Update asmtools version to 9.0.b14


theresa-m:
The latest version of asmtools 8.0 includes support to build jasm files with the early laval frame introduced in JEP draft: Strict Field Initialization in the JVM. I would like to update the asmtools version in this project to test this feature in OpenJ9.

I see that the link can be updated in [getDependencies.pl](https://github.com/adoptium/TKG/blob/master/scripts/getDependencies.pl#L127-L132) however I'm not sure how to add the new version to the Adoptium dependency pipeline. Happy to make the change if pointed in the right direction.

PR #772, opened 2025-11-13
Update asmtools to 9.0.b14

theresa-m:
Fixes: https://github.com/adoptium/TKG/issues/770


karianna (approved):

smlambert:
Is this needed for January?  We have already tagged for Jan 2026 releases, so if it is needed, we will have to update the tag for TKG in the aqa-tests/testenv/testenv.properties file.

SL/edit: I think we will need to update the tag anyway, to pick up some other recent changes in this repo.

PR #838, opened 2026-05-26
Fix JDK26 jtreg version mapping to jtreg_8_2_1_1

annaibm:
- Change JDK26 mapping from jtreg_8_1_1 to jtreg_8_2_1_1 to match the requirement in TEST.ROOT.

### DIFF
--- a/scripts/getDependencies.xml
+++ b/scripts/getDependencies.xml
@@ -47,14 +47,10 @@
 		<condition property="jtregTar" value="jtreg_7_4_1">
 			<matches pattern="^24$" string="${JDK_VERSION}"/>
 		</condition>
-		<!-- version 25 -->
-		<condition property="jtregTar" value="jtreg_8_2">
-			<matches pattern="^25$" string="${JDK_VERSION}"/>
-		</condition>
-		<!-- version 26 -->
-		<condition property="jtregTar" value="jtreg_8_1_1">
-			<matches pattern="^26$" string="${JDK_VERSION}"/>
-		</condition>
+		<!-- versions 25, 26 -->
+        <condition property="jtregTar" value="jtreg_8_2_1_1">
+            <matches pattern="^(25|26)$" string="${JDK_VERSION}"/>
+        </condition>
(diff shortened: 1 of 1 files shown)


annaibm:
Before FIX: https://hyc-runtimes-jenkins.swg-devops.com/job/Grinder/60287/ 
```
TESTING:
11:46:02  Error: The testsuite at C:\Users\jenkins\workspace\Grinder\aqa-tests\openjdk\openjdk-jdk\test\jdk requires jtreg version 8.2.1+1 or higher and this is jtreg version 8.1-dev+1.
11:46:02  -----------------------------------
11:46:02  jdk_custom_0_FAILED
11:46:02  -----------------------------------
```

annaibm:
fyi @keithc-ca

keithc-ca:
OpenJDK still only requires 8.1+1 in jdk26; see [test/jdk/TEST.ROOT](https://github.com/openjdk/jdk26u/blob/master/test/jdk/TEST.ROOT#L126).
Where do you see a requirement for 8.2.1+1?

annaibm:
The requirement I'm seeing is in ibmruntimes/openj9-openjdk-jdk on the openj9 branch: https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/test/jdk/TEST.ROOT#L150 — not in upstream OpenJDK

smlambert:
https://github.com/adoptium/TKG/pull/838#issuecomment-4547588992 and https://github.com/adoptium/TKG/pull/838#issuecomment-4547814806 indicate we are heading towards a breaking change, which has happened previously when the extensions repo is not aligned with the upstream OpenJDK repo.  In that case, I am guessing working from USE_TESTENV_PROPERTIES=true and a tagged copy of TKG that would fetch the correct version could be the course of action to take.

keithc-ca:
> The requirement I'm seeing is in ibmruntimes/openj9-openjdk-jdk on the openj9 branch:

That's the head stream (not jdk26). Java 27+ is already configured to use 8.2.1+1.

PR #847, opened 2026-06-18
Add Unicode test data and ICU4J JARS to getDependencies.pl

samyoon20:
summary:

Adds dependency entries to getDependencies.pl for the MBCS Unicode tests.

Previously, Unicode data files were stored directly in the aqa-tests git
repository. This change moves them to external dependencies downloaded
from unicode.org at build time, matching how other TKG dependencies work.

Changes:
- Add icu4j-61_1.jar and icu4j-localespi-61_1.jar (Maven Central)
- Add GB18030-2022.txt mapping file (unicode.org)
- Add Unihan.zip archives for 9 Unicode versions (10.0.0–17.0.0)
- Add individual UCD files (UnicodeData, Blocks, Scripts,
  NormalizationTest, PropertyValueAliases) for 9 Unicode versions

UnicodeData.txt and others are downloaded as individual files rather
than extracting from UCD.zip, the zip is 5.7MB compressed but we only
need 5 files (4.3MB). Unihan.zip must remain as a zip because
Unihan_IRGSources.txt is not available as a download (unsure why)

All entries use inline SHA256 checksums (shaalg => '256') so corrupted
or missing cache files are detected and re-downloaded automatically. (have not verified yet whether this is completely necessary yet)

Verified on Jenkins (Grinder): MBCS_Tests_codepoint_linux and
MBCS_Tests_unicode_linux both pass with these entries in place.

Assisted by IBM Bob

Related: adoptium/aqa-tests#5161

### DIFF
--- a/scripts/getDependencies.pl
+++ b/scripts/getDependencies.pl
@@ -288,12 +288,6 @@ my %base = (
 		sha1 => '6e4553f3b5fffe0d312df324d020ef1278d9595932ae03f4e8a2d427de83cdcd',
 		shaalg => '256'
 	},
-	unicode_unihan_12_0_0 => {
-		url => 'https://www.unicode.org/Public/zipped/12.0.0/Unihan.zip',
-		fname => 'Unihan-12.0.0.zip',
-		sha1 => '6e4553f3b5fffe0d312df324d020ef1278d9595932ae03f4e8a2d427de83cdcd',
-		shaalg => '256'
-	},
 	unicode_unihan_11_0_0 => {
 		url => 'https://www.unicode.org/Public/zipped/11.0.0/Unihan.zip',
 		fname => 'Unihan-11.0.0.zip',
--- a/scripts/getDependencies.pl
+++ b/scripts/getDependencies.pl
@@ -238,7 +238,7 @@ my %base = (
 	},
 	unicode_gb18030_2022 => {
 		url => 'https://www.unicode.org/Public/mappings/iso10646/GB18030-2022.txt',
-		fname => 'UnicodeData-GB18030-2022.txt'
+		fname => 'GB18030-2022.txt'
 	},
 	unicode_unihan_17_0_0 => {
 		url => 'https://www.unicode.org/Public/17.0.0/ucd/Unihan.zip',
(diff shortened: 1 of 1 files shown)


sophia-guo:
@samyoon20 Unicode has pre-packaged .zip under archived zipped directory https://www.unicode.org/Public/zipped/.  We can download the zip archives instead of .txt from the rwa directory.

samyoon20:
@sophia-guo updated! switched to ZIP archives from '/zipped/' directory as suggested. 
- downloads were tested and verified
- version 17.0.0 uses '/ucd/' path because I believe zipped is not yet available

sophia-guo:
https://ci.adoptium.net/view/Test_grinder/job/test.getDependency/2885/artifact/, job looks fine.

Compared with other archives seem both 9 Unihan.zip and UCD.zip are larger (6-8 MB). For unihan only one file Unihan_IRGSources is needed. For UCD 5 files are needed but UCD.zip has much more files than we needed.

samyoon20:
@sophia-guo that makes sense — my Phase 1 goal was mainly to validate that the Unicode data can be downloaded successfully in Jenkins, so I used the available packaged archives from unicode.org.
If only Unihan_IRGSources and 5 specific UCD files are actually needed for test execution, I can take a closer look at whether it would be better to narrow the downloads to just those files in a follow-up change, rather than pulling the full ZIP bundles.

smlambert:
Eventually, narrowing the downloads would be preferable, as these get transferred to every test node.

sophia-guo:
> Sam: for ex. have Unihan_IRGSources.txt file, and then the rest being zips

Yes,  I think for now having Unihan_IRGSources.txt file and the rest being zips is fine. Under currently getDependency.pl I don't think there is an easy way to partial download with expected files as zip file, which we can improve at later phase.

sophia-guo (dismissed):

samyoon20:
@sophia-guo would love some additional feedback on this!

- the getDependencies.pl should be specific to the 5 UCD files instead of downloading the larger zip file sono extraction step is needed in build.xml since the files already come
- Unihan_IRGSources.txt isn't available as a direct download (returns 404 error), only though the zip, so having a Unihan ZIP for now is unavoidable
- removed unicode_unihan_12_0_0 (identical to 12.1.0)

sophia-guo (approved):

karianna (approved):

PR #857, opened 2026-07-03
Update jtreg version to 8.3+1 for JDK17,JDK21 & JDK25

Amrutha-Kanhirathingal:
Modify jtreg version of jdk17,jdk21 &jdk25 to 8.3+1

Related : https://github.ibm.com/runtimes/automation/issues/1083

Signed-off-by : Amrutha Kanhirathingal <Amrutha.Kanhirathingal@ibm.com>

### DIFF
--- a/scripts/getDependencies.xml
+++ b/scripts/getDependencies.xml
@@ -40,7 +40,7 @@
 			<matches pattern="^(11|1[89]|2[023])$" string="${JDK_VERSION}"/>
 		</condition>
 		<!-- versions 17, 21 -->
-		<condition property="jtregTar" value="jtreg_7_5_2_1">
+		<condition property="jtregTar" value="jtreg_8_3_1">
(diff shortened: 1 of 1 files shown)


pshipton:
I'm not sure what the problem is, but I don't believe this is the solution. The original jtreg values are correct as per
https://github.com/ibmruntimes/openj9-openjdk-jdk17/blob/openj9/test/lib-test/TEST.ROOT
https://github.com/ibmruntimes/openj9-openjdk-jdk21/blob/openj9/test/lib-test/TEST.ROOT
https://github.com/ibmruntimes/openj9-openjdk-jdk25/blob/openj9/test/lib-test/TEST.ROOT

Amrutha-Kanhirathingal:
Sure @pshipton ,

Thanks for the update ,

Closing this issue since it is working by setting USE_TESTENV_PROPERTIES=true.

Issue #561, opened 2024-05-30
system test jars get redownloaded again


llxia:
When running system tests again on the same machine, the system test jars get re-redownloaded again

https://openj9-jenkins.osuosl.org/job/Grinder/3641/consoleFull
```
00:00:53.838  + perl ./aqa-tests/TKG/scripts/getDependencies.pl -path /home/jenkins/workspace/Grinder/../../testDependency/system_lib -task default -customUrl https://openj9-jenkins.osuosl.org//job/test.getDependency/lastSuccessfulBuild/artifact/
00:00:53.838  --------------------------------------------
00:00:53.838  path is set to /home/jenkins/workspace/Grinder/../../testDependency/system_lib
00:00:53.838  task is set to default
00:00:53.838  dependencyList is set to all
00:00:53.838  --------------------------------------------
00:00:53.838  Starting download third party dependent jars
00:00:53.838  --------------------------------------------
00:00:53.838  downloading dependent third party jars to /home/jenkins/workspace/Grinder/../../testDependency/system_lib
00:00:53.838  /home/jenkins/workspace/Grinder/../../testDependency/system_lib/org.eclipse.osgi-3.16.100.jar exists with correct hash, not downloading
00:00:53.838  /home/jenkins/workspace/Grinder/../../testDependency/system_lib/junit-jupiter-engine.jar exists with correct hash, not downloading
00:00:53.838  /home/jenkins/workspace/Grinder/../../testDependency/system_lib/objenesis.jar exists with correct hash, not downloading
00:00:53.838  downloading https://openj9-jenkins.osuosl.org//job/systemtest.getDependency/lastSuccessfulBuild/artifact/systemtest_prereqs/log4j/log4j-core.jar
00:00:54.311  --> file downloaded to /home/jenkins/workspace/Grinder/../../testDependency/system_lib/log4j/log4j-core.jar
00:00:54.311  Checksum verification skipped for /home/jenkins/workspace/Grinder/../../testDependency/system_lib/log4j/log4j-core.jar
00:00:54.311  downloading https://openj9-jenkins.osuosl.org//job/systemtest.getDependency/lastSuccessfulBuild/artifact/systemtest_prereqs/apache-ant/lib/ant-launcher.jar
00:00:54.783  --> file downloaded to /home/jenkins/workspace/Grinder/../../testDependency/system_lib/apache-ant/lib/ant-launcher.jar
[...]

PR #564, opened 2024-06-03
Added system test prestage jars

annaibm:
- Enables to avoid system test prerequisite zip file being empty.
- System jars are downloaded under system_lib.

related: adoptium/aqa-tests#4912
related: https://github.com/adoptium/TKG/issues/560
related: https://github.com/adoptium/TKG/issues/561

PR #2, opened 2019-10-30
Initial contribution

renfeiw:
this is the initial contribution of the Test Kit Gen project

Signed-off-by: Renfei Wang <renfeiw@ca.ibm.com>

Issue #1, opened 2019-10-08
Update openjdk-test and TSC docs once this repo is populated


smlambert:
Centralize test doc it is currently laid out across the various test locations, the 'best' overall view currently resides in the openjdk-tests/README.md file, but we should have a ToC and a description of why test material lives across locations, and how to find info in an efficient way.

karianna:
I've updated the TSC repo

smlambert:
Will close this now... as the remainder of test doc updates will be completed in openjdk-tests repo under: https://github.com/AdoptOpenJDK/openjdk-tests/issues/1558

PR #864, opened 2026-07-16
fix(getDependencies): stop system test jars from being re-downloaded on every run

samyoon20:
### Summary
This PR completes the issue adoptium/TKG#561 fix in scripts/getDependencies.pl for system test prerequisite jars.

### What changed
Added/fixed checksum metadata for the affected is_system_test jars.
Migrated the updated SHA-256-backed entries to use sha256 => ... instead of storing SHA-256 values under sha1.
Kept backward compatibility by allowing checksum lookup to use either:
sha256
legacy sha1
Restored pre-download shaurl resolution so valid existing cached artifacts can still be reused when the expected checksum is provided via checksum sidecar rather than embedded metadata.
Why
Previously, several system test prerequisite jars were missing usable pinned checksum metadata, which prevented consistent cache-hit behavior and caused unnecessary re-downloads.

**This PR ensures that:**
the affected system jars use the correct pinned SHA-256 values,
checksum naming matches the actual digest type for the updated entries,
legacy checksum entries still continue to work,
and the expected checksum can still be resolved early enough to preserve warm-run cache reuse for shaurl-based dependencies.

### DIFF
--- a/scripts/getDependencies.pl
+++ b/scripts/getDependencies.pl
@@ -751,12 +751,11 @@ if ($task eq "clean") {
 				# if the sha file exists locally, parse it and get the expected sha
 				if (-e $shafn) {
 					$expectedsha = getShaFromFile($shafn, $fn);
+				} elsif ($shaurl) {
+					downloadFile($shaurl, $shafn);
+					$expectedsha = getShaFromFile($shafn, $fn);
 				}
 			}
-			# Note: do NOT fetch shaurl here. The remote sidecar is only fetched
-			# after a download (below) so we never make a network call just to
-			# decide whether to skip. Warm-cache reuse is handled by the inline
-			# sha256/sha1 value or a locally present sidecar file.
 		}
 
 		if ($expectedsha && $digest eq $expectedsha) {
--- a/scripts/getDependencies.pl
+++ b/scripts/getDependencies.pl
@@ -709,14 +709,14 @@ if ($task eq "clean") {
 		# if url_testDependency is provided, use url_testDependency and reset $url and $shaurl
 		if ($url_testDependency ne "") {
 			if (defined $jars_info[$i]{is_system_test} && $jars_info[$i]{is_system_test} == 1) {
-				$url_testDependency =~ s/test.getDependency/systemtest.getDependency/;
+				$url_testDependency =~ s/test\.getDependency/systemtest.getDependency/;
 				$url_testDependency .= "systemtest_prereqs/";
 				$url_testDependency .= $jars_info[$i]{dir};
 			}
 
 			# Ensure exactly one slash between the base URL and the filename,
 			# regardless of whether the caller provided a trailing slash.
-			$url_testDependency =~ s/\/$//;
+			$url_testDependency =~ s/\/+$//;
 
 			$url = "$url_testDependency/$jars_info[$i]{fname}";
 
--- a/scripts/getDependencies.pl
+++ b/scripts/getDependencies.pl
@@ -707,8 +707,11 @@ if ($task eq "clean") {
 				$url_testDependency =~ s/test.getDependency/systemtest.getDependency/;
 				$url_testDependency .= "systemtest_prereqs/";
 				$url_testDependency .= $jars_info[$i]{dir};
-				$url_testDependency .= '/' unless $url_testDependency =~ /\/$/;
 			}
+
+			# Ensure exactly one slash between the base URL and the filename,
+			# regardless of whether the caller provided a trailing slash.
+			$url_testDependency =~ s/\/$//;
 
 			$url = "$url_testDependency/$jars_info[$i]{fname}";
 
--- a/scripts/getDependencies.pl
+++ b/scripts/getDependencies.pl
@@ -595,7 +595,8 @@ my %system_jars = (
 		url => 'https://repo1.maven.org/maven2/org/apache/ant/ant-launcher/1.8.1/ant-launcher-1.8.1.jar',
 		dir => 'apache-ant/lib',
 		fname => 'ant-launcher.jar',
-		sha1 => 'c99d018fcc43a1540e465b9a097508b19075198c',
+		sha1 => '28cfb70020dfdb9e4e261ceed16c43aed715c1d11390e3ddbecda1e548253168',
+		shaalg => '256',
 		is_system_test => 1
 	},
 	asm => {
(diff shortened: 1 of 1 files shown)


samyoon20:
Verification on a non-Azure Grinder agent confirms the expected behavior.

Run adoptium/TKG#1:
The affected dependencies were downloaded and SHA verification passed:

downloading .../asm/asm.jar → SHA verification passed
downloading .../log4j/log4j-api.jar → SHA verification passed
downloading .../apache-ant/lib/ant-launcher.jar → SHA verification passed
downloading .../log4j/log4j-core.jar → SHA verification passed

Run adoptium/TKG#2:
The same dependencies were found in cache and skipped:

/home/jenkins/.../asm/asm.jar exists with correct hash, not downloading
/home/jenkins/.../log4j/log4j-api.jar exists with correct hash, not downloading
/home/jenkins/.../log4j/log4j-core.jar exists with correct hash, not downloading
/home/jenkins/.../junit/hamcrest-core.jar exists with correct hash, not downloading
/home/jenkins/.../apache-ant/lib/ant-launcher.jar exists with correct hash, not downloading
/home/jenkins/.../junit/junit.jar exists with correct hash, not downloading
/home/jenkins/.../cvsclient/org-netbeans-lib-cvsclient.jar exists with correct hash, not downloading

Agent: test-ibmcloud-rhel8-x64-1 (RHEL 8, non-Azure)

This provides validation that the added SHA-1 metadata enables checksum validation and cache hits for the affected dependencies on a persistent agent.

https://ci.adoptium.net/job/Grinder/17761/ *cancelled b/c only needed parts of console output
https://ci.adoptium.net/job/Grinder/17762/ ^^

scripts/getDependencies.pl
```diff
@@ -833,8 +789,6 @@
 				print "Please delete $filename and rerun the program!";
 				die "ERROR: sha checksum error.\n";
 			}
-		} elsif ($ignoreChecksum) {
-			print "Checksum verification skipped for $filename\n";
```
karianna, line 837: We should keep this in

scripts/getDependencies.pl
```diff
@@ -595,6 +595,7 @@
 		url => 'https://repo1.maven.org/maven2/org/apache/ant/ant-launcher/1.8.1/ant-launcher-1.8.1.jar',
 		dir => 'apache-ant/lib',
 		fname => 'ant-launcher.jar',
+		sha1 => 'c99d018fcc43a1540e465b9a097508b19075198c',
```
karianna: sha-1 is deprecated, please use sha256

scripts/getDependencies.pl
```diff
@@ -704,10 +709,9 @@
 		# if url_testDependency is provided, use url_testDependency and reset $url and $shaurl
 		if ($url_testDependency ne "") {
 			if (defined $jars_info[$i]{is_system_test} && $jars_info[$i]{is_system_test} == 1) {
-				$url_testDependency =~ s/test.getDependency/systemtest.getDependency/;
+				$url_testDependency =~ s/test\.getDependency/systemtest.getDependency/;
```
karianna: Why this change

karianna (changes requested):
I have concerns that this weakens our security posture here.  I think this needs a simpler/fresh approach that simply checks if the file is already on the system (including checking the sha again) and then skipping download in that case

samyoon20:
Hi @karianna, thanks for the thorough review, really appreciate the feedback
I address the SHA-1 -> SHA-256, for the sha1 naming, a pre-existing convention? perhaps renaming is out of scope for now. I removed the pre-download shaurl network fetch. I brought back the else if ($ignoreChecksum) log, and also returned to the preserved fallback behavior for $third_party_url (from Copilot's addressed issue)

samyoon20:
@jiekang Could I get a review on this PR if you have the time!

jiekang:
@samyoon20 Can you walk me through the design here? What is intended to happen in the basic different cases? (When file exists, doesn't exist, sha exists, doesn't exist, etc.)

samyoon20:
@jiekang yes! 

there's two guards before attempting any downloads:

Guard1 (SHA match): if the file exists on the disk AND there's an expected checksum, if they match we can skip download, if they don't match we redownload

Guard2: would only kick in if there's an entry that has neither a sea-256 or shaurl. where the existing file would be accepted without verification

so the cases would be:
File exists, SHA matches → skip, print "exists with correct hash, not downloading"
File exists, SHA mismatches → re-download and re-verify
File doesn't exist, SHA known → download, then verify against expected SHA; mismatch → fatal error
File doesn't exist, SHA comes from shaurl → download the jar, then download the sha file and verify
File exists, no SHA defined at all → skip via Guard 2, no verification
File doesn't exist, no SHA defined at all → download, skip checksum verification

I also have sha256 values but the naming is sha1, and I was wondering if I should just rename everything, but was wondering if it was too much changes in this pr

karianna:
I recommend the rename to avoid future confusion

samyoon20:
@jiekang could I also get a Copilot review on this PR as well? Thank you!

scripts/getDependencies.pl
jiekang, line 1: While we're here, can we adjust all the entries with shaalg '256' to have sha256 constant instead of sha1.

For example, line 242:
https://github.com/samyoon20/TKG/blob/13d34b26f562b40317c409db41756e872fcec1fc/scripts/getDependencies.pl#L242
```
icu4j_61_1 => {
		url => 'https://repo1.maven.org/maven2/com/ibm/icu/icu4j/61.1/icu4j-61.1.jar',
		fname => 'icu4j-61_1.jar',
		sha1 => '55c98eb1838b2a4bb9a07dc36bd378532d64d0cdcb7ceee914236866a7de4464',
		shaalg => '256'
	},
```

samyoon20:
sha1 labels that were sha256 values have been renamed, and all entries with shaalg '256' now have sha256 constants

local testing confirms that for a cold run: files downloaded and the shared verification passed
warm run: outputs exists with correct hash, not downloading, and a corrupted file will trigger a redownload

grinder we checked a warm run: which each jar is hitting exists with correct hash, not downloading.

sophia-guo:
> This PR completes the issue 561 fix in scripts/getDependencies.pl for system test prerequisite jars.

Which issue is this PR linked? Did you mean adoptium/aqa-tests#7516?

samyoon20:
@sophia-guo apologies, I realized the issue wasn't linked properly! issue adoptium/TKG#561 should be it!

samyoon20:
newest grinder tests confirm

cold run: every jar is downloaded + SHA verification passed https://ci.adoptium.net/job/Grinder/17936/console

warm run: caching works and no re-downloading on repeat runs
https://ci.adoptium.net/job/Grinder/17937/console

we can ignore the make failures since it's unrelated to the PR (only needed a dynamic node to complete the cold run/warm run tests)

sophia-guo:
I think adoptium/TKG#561 was fixed by adoptium/TKG#564  https://github.com/adoptium/TKG/pull/564#issuecomment-2147627651 . See grinder https://ci.adoptium.net/job/Grinder/17930/console using TKG master branch, System test jars don't get redownloaded again , if it already exist.

scripts/getDependencies.pl
```diff
@@ -729,41 +744,38 @@
...
 				}
 			}
-
-			# if expectedsha is not set above and shaurl is provided, download the sha file
-			# and parse the file to get the expected sha
-			if (!$expectedsha && $shaurl) {
-				downloadFile($shaurl, $shafn);
-				$expectedsha = getShaFromFile($shafn, $fn);
-			}
+			# Note: do NOT fetch shaurl here. The remote sidecar is only fetched
```
sophia-guo: Not sure what does this means. For current logic, 

```
if (!$expectedsha && $shaurl) {
				downloadFile($shaurl, $shafn);
				$expectedsha = getShaFromFile($shafn, $fn);
			}
```

Should move to upper block as following to avoid the redownload the shafile.

```
if (defined $shafn && $shafn ne '') {
				$shafn = $path . $sep . $shafn;
				# if the sha file exists, parse the file and get the expected sha
				# if the sha file exists locally, parse it and get the expected sha
				if (-e $shafn) {
					$expectedsha = getShaFromFile($shafn, $fn);
				} else if ($shaurl) {
				    downloadFile($shaurl, $shafn);
				   $expectedsha = getShaFromFile($shafn, $fn);
			}		
}
```

samyoon20: will do!

samyoon20:
@sophia-guo with the confirmed fixes for the redownloads.. the PR's only contribution would be the addition of checksum for the system_jar entries, and renaming asm and mauve sha1 values to sha256, if we think that this is still a helpful update~

samyoon20:
This PR has been split to two, please reference issue #881 & issue #882 for more details!

